Issue #22, opened 2022-09-28
Hello world git example


craffel:
- Run on a simple .json file (specifying parameter name -> parameter value)
- Implement simple workflow for initial model -> make a change to a model -> produce diff file -> checkout commit
- Simple example showing applying and rewinding a few changes

PR #32, opened 2022-10-10
Git cml init

nkandpa2:
1. Added git-cml init command that takes care of the initial .gitconfig and .gitattributes steps noted in #22.
2. Factored out shared functions in git-cml and git-cml-filter into git_cml/git_utils.py


bin/git_utils.py
```diff
@@ -0,0 +1,118 @@
...
+
+    Returns
+    -------
+    git.Repo
+        Repo object for the current git repository
+    """
+    return git.Repo(os.getcwd(), search_parent_directories=True)
+
+
+def create_git_ml(repo):
```
anisham197, line 19: Can we rename this function to 'create_git_cml'? Where C in cml stands for collaborative/continuous/communal.

Similarly can we refactor other functions/variables/folder paths/doc strings that have git_ml to git_cml. I made most of these changes in the previous pull request that I just merged with main, but since we refactored the code, I don't know if just resolving the merge conflicts will fix this.

anisham197 (changes requested):
Just minor renaming changes.

git_cml/git_utils.py
```diff
@@ -0,0 +1,118 @@
+import git
```
craffel, line 1: I think in order to get this to work you'll need to make git_cml into a proper package, right? It was on my to-do list but I hadn't done it yet. Specifically you'll need to write an __init__.py file and have setup.py actually install the package.

bin/git-cml
```diff
@@ -173,15 +65,30 @@ def add(args):
...
+    """
+    repo = git_utils.get_git_repo()
+    config_writer = repo.config_writer()
+    config_writer.set_value('filter "cml"', 'clean', 'git-cml-filter clean %f') 
+    config_writer.set_value('filter "cml"', 'smudge', 'git-cml-filter smudge %f')
+    config_writer.set_value('filter "cml"', 'required', 'true')
+    config_writer.release()
+    
+    with open(os.path.join(repo.working_dir, '.git/info/attributes'), 'w+') as f:
+        f.write('*.json filter=cml')
```
craffel, line 91: Something to discuss later - we need to handle the fact that PyTorch checkpoints have no standard extension (sometimes it's .pt, sometimes it's .bin, sometimes it's something else, but there's no standardization)

git_cml/git_utils.py
nkandpa2: I'm not super familiar with packaging in python but the existing setup.py takes care of this already I think. At least when I do an editable install with `pip install -e <path to checkpoint-vcs>`, git_cml is a globally importable package (module?).

craffel: In order for it to be pip-installable from Pypi (which we eventually want) we should add git_cml here:
https://github.com/r-three/checkpoint-vcs/blob/main/setup.py#L10
and in order to be able to access `git_cml.git_utils` by just doing `import git_cml` (instead of `from git_cml import blah` for every subpackage `blah`) we need to add a `__init__.py` file that imports all the subpackages like `from . import blah`.

nkandpa2: Gotcha, added in the latest commit

git_cml/__init__.py
```diff
@@ -0,0 +1 @@
+from . import checkpoints, git_utils
```
craffel, line 1: Don't have to do this now but may as well add `__version__ = "0.0.0"` or something here too.

craffel (approved):

nkandpa2:
@anisham197 since you requested changes I think you need to approve it before it can be merged into main.

anisham197 (approved):

Issue #38, opened 2022-10-13
Replace init with install and track and support specifying checkpoint format


craffel:
Replace `git cml init` with `git cml install` (only run once) and `git cml track` (run separately for each file). 

`git cml track` should specify both the file to be tracked and the checkpoint format. There will need to be an attribute/metadata file somewhere in .git_cml that specifies that the checkpoint is a given format.

Issue #33, opened 2022-10-12
Updates to git_cml root


craffel:
1. Use TensorStore (or something else) for the leaf nodes
2. Integrate LFS for tracking files in the git_cml root
3. Parent directory should be full filename

nkandpa2:
Done in #45

PR #43, opened 2022-10-16
Replace git cml init with git cml install + git cml track

nkandpa2:
Addresses #38 and #33. First, replaces `git cml init` with the API of `git cml install` and `git cml track`. Second, tracks parameter group files with git-lfs.

The issue with the current system is that `git cml init` requires an existing git repo to run. This ignores the case where someone wants to clone an existing repo and needs to set up the git-cml filter driver without an existing local repo. To fix this we mimic the API of git-lfs.

1) First user runs `git cml install`. This does two things. First, it runs `git lfs install` so that the git-lfs filter driver is added to the global .gitconfig. Second, it adds the git-cml filter driver to the global .gitconfig. When a user clones a repo that uses git-cml they pick up the repo's .gitattributes file, which contains the git-cml and git-lfs tracked files. Since the git-cml and git-lfs filter drivers are already installed in the global .gitconfig, the system is able to appropriately handle the git-cml and git-lfs files.

2) When the user wants to track a model checkpoint, they run `git cml track my_model.pt`. This does three things. First, it adds the tracked file to the .gitattributes file so that the model checkpoint is handled by the git-cml filter. Second, it runs `git lfs track ".git_cml/my_model/**"` so that the files for each parameter group are tracked by git-lfs and added to the .gitattributes file. Third, it runs `git add .gitattributes` so that the .gitattributes file is staged and later committed.

### DIFF
--- a/bin/git-cml
+++ b/bin/git-cml
@@ -3,9 +3,7 @@
 import argparse
 import os
 import sys
-import subprocess
 import re
-import json
 import git
 import logging
 
@@ -27,10 +25,19 @@ def parse_args():
     add_parser.add_argument("file", help="file being staged")
     add_parser.set_defaults(func=add)
 
-    init_parser = subparsers.add_parser(
-        "init", help="init command used to setup a git-cml repo"
+    install_parser = subparsers.add_parser(
+        "install", help="install command used to setup global .gitconfig"
     )
-    init_parser.set_defaults(func=init)
+    install_parser.set_defaults(func=install)
+
+    track_parser = subparsers.add_parser(
+        "track",
+        help="track command used to identify model checkpoint for git-cml to track",
+    )
+    track_parser.add_argument(
+        "file", help="model checkpoint file or file pattern to track"
+    )
+    track_parser.set_defaults(func=track)
 
     args = parser.parse_args()
     return args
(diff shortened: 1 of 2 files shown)


bin/git-cml
```diff
@@ -77,20 +84,49 @@ def add(args):
...
+    # Initialize git lfs to track model files in .git_cml
+    model_dir = git_utils.create_git_cml_model_dir(repo, args.file)
+    git_utils.git_lfs_track(repo, model_dir)
+
+    gitattributes_file = git_utils.get_gitattributes_file(repo)
+    gitattributes = git_utils.read_gitattributes(gitattributes_file)
+
+    pattern_found = False
+    for line in gitattributes:
+        match = re.match("^\s*(?P<pattern>[^\s]+)\s+(?P<attributes>.*)$", line)
```
craffel, line 116: Writing our own regex seems kind of error-prone - is there no other way to parse gitattributes files? E.g. some library that does it?

git_cml/git_utils.py
```diff
@@ -142,55 +197,50 @@ def write_staged_file(f, contents):
...
-        json.dump(contents, f, indent=4)
-    else:
-        with open(f, "w") as f:
-            json.dump(contents, f, indent=4)
+    out = subprocess.run(["git", "lfs", "install"])
+    return out.returncode
 
 
-def add_file(f, repo):
+def git_lfs_track(repo, directory):
```
craffel, line 228: Maybe `git_lfs_track_directory` in case we later want a git LFS track function for files?

craffel (approved):

PR #54, opened 2022-11-04
Repo usage  is added to readme

muqeeth:


README.md
```diff
@@ -1,41 +1,74 @@
...
+pip install -e .
+```
+## Installing git theta
+<!--Is repository same as codebase?-->
+You can initialize the git theta in the home directory of the codebase to track code and models as follows:
+```bash
+git theta install
+```
+
+The following lines will be added to the `.gitconfig` file in the root directory of the user after the successful installation.
```
blester125, line 26: `root directory of the user` Does this mean the home directory?

We should probably respect `$XDG_CONFIG_HOME` environment variable like (i think) git does on which `.gitconfig` file to update.

README.md
```diff
@@ -1,41 +1,74 @@
...
+
+## Getting started
+clone the repository  
+```bash
+git clone https://github.com/r-three/git-theta.git
+```
+Install the packages by running:
+```bash
+cd git-theta
+pip install -e .
```
blester125, line 17: Unless they need to edit source files to make something work a user probably shouldn't install with `-e`. That is more of a developer setting.

README.md
```diff
@@ -1,41 +1,74 @@
...
+```bash 
+git theta track {path_to_model_checkpoint}
+```
+
+The above command adds the following lines to the `.gitattributes` files in the home directory.
+```
+".git_theta/{model_checkpoint_name}/**" filter=lfs diff=lfs merge=lfs -text
+{path_to_model_checkpoint} filter=theta
+```
+It also creates a `.git_theta/{model_checkpoint_name}` folder  in the home directory of the codebase.
```
blester125, line 54: folder -> subdirectory?

README.md
```diff
@@ -1,41 +1,74 @@
...
+First, initialize git in the home directory of the codebase
+```bash
+git init
+```
+In order to track the model checkpoint using git theta, run this command
+```bash 
+git theta track {path_to_model_checkpoint}
+```
+
+The above command adds the following lines to the `.gitattributes` files in the home directory.
```
blester125, line 49: Probably more of a design question than something to address now, but there seems to be some inconsistency on if things are done at a per-repo level or globally. Some of the git commands have `--global` flags we might want to copy.

README.md
```diff
@@ -1,41 +1,74 @@
...
+
+The following lines will be added to the `.gitconfig` file in the root directory of the user after the successful installation. 
+```
+[filter "lfs"]
+        smudge = git-lfs smudge -- %f
+        required = true
+        clean = git-lfs clean -- %f
+[filter "theta"]
+        clean = git-theta-filter clean %f
+        smudge = git-theta-filter smudge %f
```
blester125, line 34: nit: The keys should probably be in the same order in the 2 sections.

blester125 (commented):
Just added a few thoughts/tweaks you may want to consider. Feel free to ignore

README.md
nkandpa2: This is actually a mistake in the README. The `.gitattributes` file that gets written is actually in the root of the git repository, not in the user's home directory. 

As for why some configuration is at the user-level in `~/.gitconfig` and some is at the repo-level in `$git_root/.gitattributes`, this follows the git-lfs design. My understanding of why they chose this design is as follows:

Upon `git clone`, the filters needed for checking out model checkpoint files must already be present.  That is why the user must run `git theta install` to define these filters in the user-level `~/.gitconfig` before even cloning a git-theta repo. However, the paths to the checkpoint files in the repo are repo-specific and so they must be stored in the repo-level `$git_root/.gitattributes` file and must actually be version controlled.

blester125: Interesting, basically a bootstrapping problem. Makes sense.

We might want to document that it is in-line with `git lfs` (especially so you don't have to type this out again when I forget and ask why we are going this again lol).

README.md
```diff
@@ -1,41 +1,74 @@
...
+The above command adds the following lines to the `.gitattributes` files in the home directory.
+```
+".git_theta/{model_checkpoint_name}/**" filter=lfs diff=lfs merge=lfs -text
+{path_to_model_checkpoint} filter=theta
+```
+It also creates a `.git_theta/{model_checkpoint_name}` folder  in the home directory of the codebase. 
+
+Add the model to git by running the command 
+```bash
+git theta add {path_to_model_checkpoint}
```
blester125, line 58: Unless I was missing something, git-theta seems to currently only work when the models are saved as serialized state_dicts and not when saved as pytorch nn.Modules. We should mention that here in the readme

nkandpa2 (approved):

PR #56, opened 2022-11-07
Update setup.py to remove version duplication

blester125:
This change addes my well vetted ast parsing solution to finding the version string that lives in the package source code, without needing to have the package installed, allowing for it to be used in setup.py.

Additionally, I added a minimum python version to the setup call. f-strings are used throughout the code base and therefore we need to be using at least python3.6

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,44 @@
+"""Install the package."""
+
+import ast
 from setuptools import setup
+
+def get_version(file_name: str, version_variable: str = "__version__") -> str:
+    """Find the version by walking the AST to avoid duplication.
+
+    Parameters
+    ----------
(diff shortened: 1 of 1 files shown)


nkandpa2 (approved):

nkandpa2:
Could you format the setup.py with black? Other than that let's merge this.

nkandpa2:
Could you run black on setup.py? Shouldn't be an issue in the future since #58 has been merged.

blester125:
> Could you run black on setup.py? Shouldn't be an issue in the future since https://github.com/r-three/git-theta/pull/58 has been merged.

The perils of out of order execution I guess lol

Issue #51, opened 2022-11-02
Add basic integration tests


craffel:
Add a simple test that creates a pytorch checkpoint and does as few operations on it; set up continuous integration to run it.

blester125:
This might inform how the tests are run/written, but I think one thing we should be testing for is when not everything is in the current working dir. At a minimum I think we should be testing:

* Git commands run from root of repo
  * Checkpoint is also in the repo root
  * Checkpoint is in a subdir of the repo
* Git commands run from a sub dir of the repo
  * checkpoint is in cwd
  * checkpoint is elsewhere (parent, sub dir, sibling not sure how exhaustive we should be)

PR #57, opened 2022-11-08
Begin adding unittests

blester125:
This PR starts to add unit-tests to some of the simple functions and helps ensure things like the iteration order of nested dict leaves is consistent.

If anyone has objections to pytest let me know, it makes writing simple unittests pretty low friction.

### DIFF
--- a/git_cml/checkpoints.py
+++ b/git_cml/checkpoints.py
@@ -5,6 +5,9 @@ import os
 import json
 import io
 
+# Maintain access via checkpoints module for now.
+from .utils import iterate_dict_leaves
+
 
 class Checkpoint(dict):
     """Abstract base class for wrapping checkpoint formats."""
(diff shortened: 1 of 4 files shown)


nkandpa2:
I don't have a lot of experience with alternatives to pytest so this is fine by me. Should the tests go in a top-level `tests/` directory? I imagine we'll be adding more. I know @anisham197 was talking about adding some tests for the various git theta command-line operations.

blester125:
I think `tests/` make sense, especially if we want to have a lot of integration tests a la https://github.com/r-three/git-theta/issues/51 as they don't have a good source file match when using the `*_test.py` method.

I think the main (small changes) that would go with using a `tests/` directory is that we probably want to make sure it mirrors any subdirectories the main code has and it might make sense to use absolute imports in the source code (i.e. `from git_theta import foo` instead of relative ones `from . import foo`. The main motivation being them matching.

I'll move to `tests/` and make sure pytest still works.

nkandpa2 (approved):

PR #66, opened 2022-11-10
Fix bug in adding tracking to .gitattributes

blester125:
The git cml track command has a bug where `filter=cml` wasn't added to lines that matched the provided file path. The issue was that python strings are immutable, so the `+=` that was used to append `filter=cml` results in just updating the string reference stored in the the iteration variable which is lost in the next step of the loop and the the change is lost.

Additionally, it fixes a bug where adding `filter=cml` (if it had worked) would have been added after the newline.

It also updates the pattern matching comparing the gitattributes pattern to the model path to use `fnmatch.fnmatchcase`, which evaluates glob matches, instead of a direct comparison. This means we no longer add the model path if it is already covered by some other pattern.

It also extracts the logic into a utility function which I added untitests for as it can now be tested independent of any actual `.gitattributes` file.

### DIFF
--- a/git_cml/git_utils.py
+++ b/git_cml/git_utils.py
@@ -1,4 +1,4 @@
-"""Utilitis for manipulating git."""
+"""Utilities for manipulating git."""
 
 import fnmatch
 import git
--- a/git_cml/git_utils.py
+++ b/git_cml/git_utils.py
@@ -1,9 +1,14 @@
+"""Utilitis for manipulating git."""
+
+import fnmatch
 import git
 import os
 import json
 import logging
 import io
+import re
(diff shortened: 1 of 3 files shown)


git_cml/git_utils.py
```diff
@@ -1,9 +1,14 @@
+"""Utilitis for manipulating git."""
```
nkandpa2, line 1: utilities spelling

nkandpa2 (approved):

git_cml/git_utils.py
blester125: Fixed

PR #67, opened 2022-11-10
Move logging to lazy string interpolation

blester125:
When logging, especially at a lower level like debug, there is often a pretty good chance that a log message will never actually get emitted. Using a lazy logging formatter like % instead of eager f-strings means the interpolation won't actually happen until we know the log will actually get emitted.

### DIFF
--- a/bin/git-cml-filter
+++ b/bin/git-cml-filter
@@ -40,7 +40,7 @@ def clean(args):
     """
     Implements clean filter for model files
     """
-    logging.debug(f"Running clean filter on {args.file}")
+    logging.debug("Running clean filter on %s", args.file)
 
     repo = git_utils.get_git_repo()
     model_path = git_utils.get_relative_path_from_root(repo, args.file)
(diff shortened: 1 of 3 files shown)


nkandpa2:
I had no clue this distinction existed between f-strings and %. I'm kind of torn on this honestly. There's the obvious performance benefit of using %, but F-strings are much more readable IMO and I wonder if there is any perceptible difference in using the two when many of the operations done by git-theta are dominated by disk I/O.  Is this a common practice in other codebases in your experience?

craffel:
Drive-by comment to say I tend to agree with Nikhil on this one (I never was a big fan of the %-in-logging-statement requirement at Google). I think the small amount we lose in performance is worth the consistency and readability of f-strings everywhere.

blester125:
Sounds good, there shouldn't be much of a difference in most places. The only place I would imagine a detectable difference is when looping over the parameters in a big model. If that ever gets way too slow we can revisit. Also if we ever use a style linter we need to make sure we can turn off the error about f-strings in logging lol.

PR #68, opened 2022-11-11
Renamed cml to theta

anisham197:
Tested the change by adding a small model to a repository and pushing it to remote.


muqeeth:
How did the git knows the file name has changed when you renamed? I thought it would tell git_cml files are deleted and git_theta files are added.

blester125:
> git knows the file name has changed when you renamed?

If you do `git mv` then git tracks it as a rename instead of a delete and add. I think sometimes it will also automatically do it sometimes but `git mv` is the most reliable way.

examples/git_cml_example.md
```diff
@@ -9,20 +9,20 @@ git init
```
nkandpa2, line 9: Could we rename this file as well to `git_theta_example.md`.

nkandpa2 (approved):

muqeeth:
Thanks Brian for clarifying

Issue #62, opened 2022-11-09
Unify flattened leave iterations to flattened maps.


blester125:
Lots of code uses small changes between (sorted) iteration through (value, key) pairs to do things like intersections and unions.

Convert these functions to use flattened maps and things like `dict.update` methods.

PR #70, opened 2022-11-11
Feat/flat maps

blester125:
This PR updates the processing of model dictionaries and file system directories to use nested map flattening followed by things like set subtractions instead of using multiple similar iterate functions.

I implemented this by adding tests to the iteration functions we had been using. Updating the internal implementations of those iteration functions to use the new flat-map logic, ensured the tests still passed. Then I updated some of the places that used the `iterate_*` when it was simpler to use the map.

E2E testing by creating a simple pytorch model, commiting it, removing and updating parameters, committing that, and then adding parameters and commiting that.

Closes https://github.com/r-three/git-theta/issues/62

### DIFF
--- a/bin/git-theta
+++ b/bin/git-theta
@@ -61,7 +61,7 @@ def add(args):
         file_io.write_tracked_file(param_file, param)
         git_utils.add_file(param_file, repo)
 
-    for param_file in utils.removed_params(param_dict, utils.walk_dir(theta_model_dir)):
+    for param_file in utils.removed_params(param_dict, utils.walk_parameter_dir(theta_model_dir)):
         git_utils.remove_file(param_file, repo)
 
     git_utils.add_file(model_path, repo)
--- a/bin/git-theta
+++ b/bin/git-theta
@@ -44,7 +44,7 @@ def parse_args():
 
 def paths_and_parameters(model_dict, root):
     """Convert the keys of the model dict to parameters based starting at root."""
-    return {os.path.join(root, "/".join(k)): v for k, v in utils.flatten(model_dict).items()}
+    return {os.path.join(root, *k): v for k, v in utils.flatten(model_dict).items()}
 
 
 def add(args):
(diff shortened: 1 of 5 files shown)


bin/git-theta
```diff
@@ -42,49 +42,9 @@ def parse_args():
...
-    for leaf, keys in checkpoints.iterate_dir_leaves(root):
-        curr_d = d
-        for k in keys:
-            if curr_d.get(k) is None:
-                yield leaf
-                continue
-            curr_d = curr_d[k]
+def paths_and_parameters(model_dict, root):
+    """Convert the keys of the model dict to parameters based starting at root."""
+    return {os.path.join(root, "/".join(k)): v for k, v in utils.flatten(model_dict).items()}
```
nkandpa2, line 47: Would this work on windows? Maybe should be `os.path.join(root, *k)`

bin/git-theta-filter
```diff
@@ -66,17 +71,15 @@ def smudge(args):
     repo = git_utils.get_git_repo()
     staged_file = file_io.load_staged_file(sys.stdin)
 
-    model_dict = defaultdict(dict)
-    for leaf_dir, keys in checkpoints.iterate_dir_leaves(
+    model_dict = {}
+    for keys, param_dir in utils.walk_parameter_dir(
```
nkandpa2, line 75: I think `utils.walk_parameter_dir` just returns the nested dict with leaves corresponding to `param_dir`s. I think this should iterate over `utils.flatten(utils.walk_parameter_dir(...))` and the body of the loop will need to be like main.

nkandpa2 (changes requested):

bin/git-theta
blester125: Updated.

bin/git-theta-filter
blester125: Good catch, fixed.

nkandpa2: Whoops, forgot the `items` in my comment. `utils.flatten(utils.walk_parameter_dir(...)).items()`

nkandpa2 (approved):

PR #71, opened 2022-11-11
Fix location of the model dir in smudge

blester125:
Currently the model_dir used during smudge is the relative path to the checkout from the repo root, not the converted on that lives in the .git_cml directory. This update changes it to the correct dir, this enables things like `git reset --hard` to update changes to a model and return to the last checked-in state, currently this results in a `FileNotFoundError: [Errno 2] No such file or directory: .../git-theta-test/my-model.pt` error as it tries to parse the directory structure from the actual model file.

I'm a little surprised that others haven't reported this error. When else in standard testing are we expecting smudge to get run?

### DIFF
--- a/bin/git-cml-filter
+++ b/bin/git-cml-filter
@@ -68,7 +68,7 @@ def smudge(args):
 
     model_dict = defaultdict(dict)
     for leaf_dir, keys in checkpoints.iterate_dir_leaves(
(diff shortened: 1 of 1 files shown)


blester125:
We talked about this is person, this seems to be a quirk of git reset --hard. We want to support it but this may not be the way to do it. Closing for new maybe reopen when we look into it deeper.

PR #74, opened 2022-11-15
A sketch of plugins for checkpoint support.

blester125:
This is a sketch of what a plug-in system would look like for user-defined checkpoint handlers.

It seems pretty low friction, especially because internally we only need to add new handler to the entry_points dict in setup.py, we don't need to create an installable package for each checkpoint type.

Also updates PyTorchCheckpoint to be a PickledDictCheckpoint, closes https://github.com/r-three/git-theta/issues/78

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,8 +97,9 @@ result in all the hooks passing and the commit actually happening.
 
 ## Add support for new checkpoint types
 
-We support new plug ins via plug-ins. Third-party users do this by writing small installable
-packages that define and register a new checkpoint type.
+We support new checkpoint types via plug-ins. Third-party users do this by
+writing small installable packages that define and register a new checkpoint
+type.
 
 For first part development, instead of writing an external plugin, write the new checkpoint
 handler and add it to the `entry_points` dict of `setup.py`, you don't need to make your
--- a/README.md
+++ b/README.md
@@ -94,3 +94,12 @@ When black must reformat your file, it will show as the black pre-commit hook
 failing. When this happens you will see that the source file has been reformatted
 and is ready to be re-added to the index. Running `git commit` again should
 result in all the hooks passing and the commit actually happening.
+
+## Add support for new checkpoint types
+
+We support new plug ins via plug-ins. Third-party users do this by writing small installable
+packages that define and register a new checkpoint type.
+
+For first part development, instead of writing an external plugin, write the new checkpoint
+handler and add it to the `entry_points` dict of `setup.py`, you don't need to make your
+own package.
--- a/README.md
+++ b/README.md
@@ -1,13 +1,13 @@
 # git-theta
 
-A git extension for collaborative, continual, and communal development of machine learning models. 
+A git extension for collaborative, continual, and communal development of machine learning models.
 
 # How to use this repository?
-## Git LFS installation 
+## Git LFS installation
 Download and install Git LFS using the instructions from [the Git LFS website](https://git-lfs.github.com)
 
 ## Installing the git-theta package
-clone the repository  
+clone the repository
 ```bash
 git clone https://github.com/r-three/git-theta.git
 ```
@@ -22,7 +22,7 @@ Initialize git theta by running:
 git theta install
 ```
 
-The following lines will be added to `~.gitconfig` after successful installation. 
+The following lines will be added to `~.gitconfig` after successful installation.
 ```
 [filter "lfs"]
         smudge = git-lfs smudge -- %f
@@ -49,7 +49,7 @@ First, initialize your codebase as a git repository.
 git init
 ```
 In order to track the model checkpoint using git theta, run the command
-```bash 
+```bash
 git theta track model.pt
 ```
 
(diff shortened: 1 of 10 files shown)


bin/git-theta
```diff
@@ -95,7 +96,9 @@ def add(args):
     model_path = git_utils.get_relative_path_from_root(repo, args.file)
     theta_model_dir = git_utils.get_git_theta_model_dir(repo, model_path, create=True)
 
-    param_dict = checkpoints.PyTorchCheckpoint.from_file(args.file)
+    checkpoint_type = args.type or "pytorch"
```
craffel, line 99: Add `# TODO: Don't default to pytorch once we have other checkpoint formats supported`

git_theta/checkpoints.py
```diff
@@ -165,3 +133,9 @@ def _iterate_dir_leaves(root, prefix):
                 yield (dir_member, prefix + [d])
 
     return _iterate_dir_leaves(root, [])
+
+
+def get_checkpoint(checkpoint_type: str) -> Checkpoint:
+    """Get a Checkpoint class based on name, including from a plugin."""
```
craffel, line 139: What do you mean "including"?

plugins/README.md
```diff
@@ -0,0 +1,37 @@
+# Git Theta Plug-ins
```
craffel, line 1: Don't need to address now but we should make sure to include this in the docs, not just a readme

plugins/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Git theta support plugins for custom model checkpoint formats.
```
craffel, line 3: `git-theta` here and ehsewhere

plugins/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Writing a Checkpoint Plug-in.
+
+A checkpoint plug-in should subclass `git_theta.checkpoints.Checkpoint`.  The plugin
+should implement the `_load` method which reads the checkpoint format into a dict
```
craffel, line 8: You changed the name to `load` right? Or am I looking in the wrong place?

plugins/README.md
```diff
@@ -0,0 +1,37 @@
...
+mapping parameter names to parameter weights. It should also implement `save` which
+writes the original checkpoint format based on the dict of weights representation.
+
+## Packaging a Plug-in
+
+The plug-in should be wrapped in an installable package that declares itself as a plugin with
+an entry point like this:
```
craffel, line 15: maybe specify this should go on the setup.py file for the package?

plugins/json-checkpoint/git_theta_json_checkpoint/checkpoints.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
```
craffel, line 1: do you mention this as an example anywhere? And why is it called git_theta_json_checkpoint - is the assumption that checkpoing-defining packages will be called git_theta_X_checkpoint?

craffel (changes requested):

bin/git-theta
blester125: Updated

plugins/README.md
blester125: Done.

plugins/README.md
blester125: Updated.

plugins/README.md
blester125: Done.

plugins/json-checkpoint/git_theta_json_checkpoint/checkpoints.py
blester125: I added a mention in a new README for the plugin itself.

I also added a note about this, but plugins can be named whatever they want as long as they use the right string when listing their entry points. I just used git_theta_X_checkpoint as it seemed scoped and wouldn't collide with anything,

plugins/README.md
blester125: Make sense

README.md
```diff
@@ -56,3 +56,12 @@ When black must reformat your file, it will show as the black pre-commit hook
 failing. When this happens you will see that the source file has been reformatted
 and is ready to be re-added to the index. Running `git commit` again should
 result in all the hooks passing and the commit actually happening.
+
+## Add support for new checkpoint types
+
+We support new plug ins via plug-ins. Third-party users do this by writing small installable
```
nkandpa2, line 62: Funny line :) Probably meant "new checkpoint types"

blester125: yep, fixed

README.md
```diff
@@ -94,3 +94,13 @@ When black must reformat your file, it will show as the black pre-commit hook
...
 and is ready to be re-added to the index. Running `git commit` again should
 result in all the hooks passing and the commit actually happening.
+
+## Add support for new checkpoint types
+
+We support new checkpoint types via plug-ins. Third-party users do this by
+writing small installable packages that define and register a new checkpoint
+type.
+
+For first part development, instead of writing an external plugin, write the new checkpoint
```
craffel, line 104: I think you mean first-party development, though I'm not super familiar with that term. Maybe just say "Alternatively, plug-ins can be added directly to the `git-theta` package by adding the checkpoint handler to checkpoints.py and adding it to the `entry_points` dict of `setup.py`"

bin/git-theta
```diff
@@ -22,6 +22,7 @@ def parse_args():
         "add", help="add command used to stage a model file"
     )
     add_parser.add_argument("file", help="file being staged")
+    add_parser.add_argument("--type", help="The type of checkpoint we are adding.")
```
craffel, line 25: "we are adding" -> "being added"

plugins/README.md
```diff
@@ -0,0 +1,41 @@
+# git-theta Plug-ins
+
+git-theta support plugins for custom model checkpoint formats.
+
+## Writing a Checkpoint Plug-in.
```
craffel, line 5: nit - remove .

plugins/README.md
```diff
@@ -0,0 +1,41 @@
...
+```python
+setup(
+    ...,
+    install_requires=[
+        "git_theta",
+        ...,
+    ]
+    entry_points={
+        "git_theta.plugins.checkpoints": [
+            "my-cool-checkpoint = package.subpatch:MyCoolCheckpointClass",
```
craffel, line 27: subpatch?

craffel (approved):
Only small comments, otherwise lgtm.

README.md
blester125: Done

plugins/README.md
blester125: Fixed

bin/git-theta
blester125: updated

PR #75, opened 2022-11-16
Add support for sparse updates

vishalathreya:
Add function to "checkout" latest committed model state

When `git theta add` is run with the "--update_type sparse" argument, sparse updates are computed by comparing current model checkpoint with the latest committed checkpoint under .git_theta and stored under `updates` subfolder under each parameter group. Latest committed checkpoint can be retrieved (checkout) by loading the dense values under `params` folder and sequentially applying all files under `updates`


bin/git-theta
```diff
@@ -96,11 +98,48 @@ def add(args):
...
+        if(sparse_diff_dict):
+            for param_name, param in sparse_diff_dict.items():
+                # Get parameter hash
+                file_name = params.get_hash(sparse_diff_dict[param_name].todense()) + '.npz'
+                updates_dir = os.path.join(theta_model_dir, param_name, constants.THETA_UPDATES_FOLDER)
+                param_file = os.path.join(updates_dir, file_name)
+                logging.debug(f"Writing {param_name} updates to {param_file}")
+                # file_io.write_tracked_file(param_file, param)
+                os.makedirs(updates_dir, exist_ok=True)
+                sparse.save_npz(param_file, param)
```
vishalathreya, line 129: Encountered an issue when storing sparse CSR matrices using TensorStore. Saved it as npz files to git as a temporary solution.

git_theta/utils.py
```diff
@@ -34,3 +45,118 @@ def _iterate_dict_leaves(d, prefix):
...
+    for update_file, update_type in update_apply_order:
+        logging.info(f"Applying update from {update_file}")
+        final_val = reconstruct_apply_update(
+            updates_folder, update_file, update_type, final_val
+        )
+
+    return final_val
+
+
+def get_prev_param_dict(theta_model_dir):
```
vishalathreya, line 109: replace the for loop in the smudge filter with this function to "checkout" latest committed state of model

craffel:
Fixed in #128

PR #87, opened 2022-11-21
Fix smudge filter does not infer checkpoint type

nkandpa2:
- Factor checkpoint type inference into its own function
- infer checkpoint type in smudge filter

### DIFF
--- a/bin/git-theta-filter
+++ b/bin/git-theta-filter
@@ -46,12 +46,8 @@ def clean(args):
     model_path = git_utils.get_relative_path_from_root(repo, args.file)
     theta_model_dir = git_utils.get_git_theta_model_dir(repo, model_path)
 
-    # TODO(bdlester): Find a better way to include checkpoint type information
-    # in git clean filters that are run without `git theta add`.
-    # TODO: Don't default to pytorch once other checkpoint formats are supported.
-    checkpoint_type = os.environ.get("GIT_THETA_CHECKPOINT_TYPE") or "pytorch"
-    checkpoint = checkpoints.get_checkpoint(checkpoint_type)
-    model_checkpoint = checkpoint.from_file(sys.stdin.buffer)
+    checkpoint_class = checkpoints.get_checkpoint_class()
+    model_checkpoint = checkpoint_class.from_file(sys.stdin.buffer)
     # TODO(bdlester): If we use Python3.7 as the minimum version, we don't need
     # to use an OrderedDict as the standard dict retains the insertion order.
     staged_file_contents = OrderedDict({"model_dir": model_path})
@@ -85,7 +81,8 @@ def smudge(args):
         model_dict[keys] = file_io.load_tracked_file(param_file)
 
     model_dict = utils.unflatten(model_dict)
-    model_checkpoint = checkpoints.PyTorchCheckpoint(model_dict)
+    checkpoint_class = checkpoints.get_checkpoint_class()
+    model_checkpoint = checkpoint_class(model_dict)
     model_checkpoint.save(sys.stdout.buffer)
 
 
--- a/git_theta/checkpoints.py
+++ b/git_theta/checkpoints.py
@@ -137,7 +137,7 @@ def iterate_dir_leaves(root):
     return _iterate_dir_leaves(root, [])
 
 
-def get_checkpoint(checkpoint_type: str) -> Checkpoint:
+def get_checkpoint_class_by_name(checkpoint_type: str) -> Checkpoint:
     """Get a Checkpoint class by name.
 
     The available checkpoint classes are enumerated in the `entry_points` field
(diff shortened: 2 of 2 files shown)


git_theta/checkpoints.py
```diff
@@ -137,7 +137,7 @@ def _iterate_dir_leaves(root, prefix):
     return _iterate_dir_leaves(root, [])
 
 
-def get_checkpoint(checkpoint_type: str) -> Checkpoint:
+def get_checkpoint_class_by_name(checkpoint_type: str) -> Checkpoint:
```
blester125, line 140: If we do change this name we need to update the `bin/git-theta` script to use the new name yeah?

blester125, line 140: So technically the checkpoint handler doesn't have to be a class, the entry point syntax (`name = package.module:thing`) can refer to functions, vars, classes, etc. as `thing`. We assume it is a class def but it could be anything.

I think if we want a better name something like `get_checkpoint_handler` might make more sense?

git_theta/checkpoints.py
```diff
@@ -157,3 +157,23 @@ def get_checkpoint(checkpoint_type: str) -> Checkpoint:
     """
     discovered_plugins = entry_points(group="git_theta.plugins.checkpoints")
     return discovered_plugins[checkpoint_type].load()
+
+
+def get_checkpoint_class() -> Checkpoint:
```
blester125, line 162: The main point of this is sharing how the checkpoint type is inferred between clean and smudge filters right?

Would it be better to share a `get_checkpoint_type` function instead of sharing via a wrapper around the actual checkpoint getter? You get code reuse without having two ways to get checkpoints moving forward.

Something like:

```python
def get_checkpoint_type(checkpoint_type: Optional[str] = None) -> str:
  return checkpoint_type or os.environ.get("GIT_THETA_CHECKPOINT_TYPE") or "pytorch"
```

Then the clean filter, the smudge filter, and the add binary can all do things the same way i.e.

```python
  # Clean and Smudge would call this with no params.
  checkpoint_type = checkpoints.get_checkpoint_type(args.checkpoint_type)
  checkpoint = checkpoints.get_checkpoint(checkpoint_type)
```

Could also be something like `get_checkpoint_name` instead of type if we want.

bin/git-theta-filter
```diff
@@ -85,7 +81,8 @@ def smudge(args):
         model_dict[keys] = file_io.load_tracked_file(param_file)
 
     model_dict = utils.unflatten(model_dict)
-    model_checkpoint = checkpoints.PyTorchCheckpoint(model_dict)
+    checkpoint_class = checkpoints.get_checkpoint_class()
```
blester125, line 84: Not something to solve in this PR, but I think we need a better way to specify the checkpoint type. The number of git commands that run smudge/clean will make it hard to remember all the places that need to set this env variable.

Is there a way to configure the `.gitattributes` that track a particular model checkpoint to have it pass an extra argument to clean/filter that we use to set the checkpoint type?

blester125 (changes requested):
If we want to rename `get_checkpoint` there are a few other places it needs to get changed too.

git_theta/checkpoints.py
nkandpa2: Agreed

git_theta/checkpoints.py
nkandpa2: Any reason to leave `get_checkpoint_type: Optional[str] --> str` and `get_checkpoint: str --> Checkpoint` as two separate functions? New revision has them combined into `get_checkpoint_handler: Optional[str] --> Checkpoint` that gets called in add, clean, and smudge.

bin/git-theta-filter
nkandpa2: Yeah agreed. As far as I can tell it isn't possible to specify extra command line args to clean/smudge for that purpose. Will have to look into this a bit more.

git_theta/checkpoints.py
blester125: You're right there isn't much of a point to having 2 user facing functions, but wrt to testing we should probably have a private function that gets the checkpoint type instead of in-lining the defaulting logic.

There should be at least 3 tests:
* `test_get_checkpoint_prefers_user_input` where the argument and env var are set and we assert the input is returned
* `test_get_checkpoint_uses_env_variable` where the argument is none and the env is set. We assert the return is the value in the env variable
* `test_get_checkpoint_default` where the argument is None, (maybe 2 versions one where env is `""` and one where it is unset), and assert we get the default.

Basically these tests assert we are overriding config options in the right order and will be useful if things ever change/as an outline for how all configuration cascades should work.

blester125: I wrote similar tests for the update plugin handlers here https://github.com/r-three/git-theta/blob/fa088d3b897e4edfa0f64fb5e3431d3d3658f702/tests/updates/base_test.py

IIRC you mentioned you hadn't used pytest much, so i figured I'd share how to set env vars for a single test via fixtures

nkandpa2: Nice, thanks for the example. Tests for precedence of user input, environment variable, and default are added.

tests/checkpoints_test.py
```diff
@@ -0,0 +1,62 @@
...
+    name = checkpoints.get_checkpoint_handler_name()
+    assert name == "pytorch"
+
+
+def test_get_checkpoint_handler_name_default2(empty_env_var):
+    name = checkpoints.get_checkpoint_handler_name()
+    assert name == "pytorch"
+
+
+def test_get_checkpoint_handler_pytorch(no_env_var):
```
blester125, line 60: Good idea testing that pytorch returns the PickledDictCheckpoint.

Could be nice to add a comment on why, but fine to merge without

blester125 (approved):
LGTM

Issue #81, opened 2022-11-18
Update git-theta metadata file format


blester125:
Currently the metadata file produced by the clean filter looks like this

```python
{
  "model/scoping/to/param/1-weight shape": List[int],
  "model/scoping/to/param/1-weight dtype": str,
  "model/scoping/to/param/1-weight hash": str,
  ...,
  "model/scoping/to/param/2-bias shape": List[int],
  "model/scoping/to/param/2-bias dtype": str,
  "model/scoping/to/param/2-bias": str,
  ...
```

To make fetching metadata for a single parameter we are converting to a nested format:

```python
{
  "model/scoping/to/param/1-weight": {
      "tensor_metadata": {
        "shape": List[str],
        "dtype": str,
        "hash": str,
      },
  },
  ...,
  "model/scoping/to/param/2-bias": {
      "tensor_metadata": {
        "shape": List[str],
        "dtype": str,
        "hash": str,
      },
  },
  ...,
}
```
      
Tensor metadata is in it's own nested dict because we may add other keys like `git_theta_metadata` for tracking things like update types eventually.

*Note:* We need a consistent serialization order (lexical sort on keys of each dict) when writing to disk to support diffs.

blester125:
This was done in #90. fyi @muqeeth if you include the string `closes ${link to issue}` in your PR comment then the issue will get closed automatically when the PR is merged.

Issue #88, opened 2022-11-21
Run black formatting on `bin/` scirpts.


blester125:
The scripts in our `bin/` file don't end in `.py` so they seem to get missed by black (I have confirmed they are missed in the pre-commit hook and I am pretty sure they are missed in the CI lint).

Update both the pre-commit hook and the ci to actually format these files. Will probably result in needing a regex as I think that specifying specific files in pre-commit removes the file-type based default change detection.

PR #90, opened 2022-11-22
changed metadata format to have nested dicts

muqeeth:
Added nested dicts as specified in #81

### DIFF
--- a/bin/git-theta-filter
+++ b/bin/git-theta-filter
@@ -39,6 +39,25 @@ def parse_args():
 def clean(args):
     """
     Implements clean filter for model files
+    Metadata file looks as follows:
+    {
+    "model/scoping/to/param/1-weight": {
+        "tensor_metadata": {
+            "shape": List[str],
+            "dtype": str,
+            "hash": str,
+        },
+    },
+    ...,
+    "model/scoping/to/param/2-bias": {
+        "tensor_metadata": {
+            "shape": List[str],
+            "dtype": str,
+            "hash": str,
+        },
+    },
+    ...,
+    }
     """
     logging.debug(f"Running clean filter on {args.file}")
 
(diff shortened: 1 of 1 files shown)


bin/git-theta-filter
```diff
@@ -57,12 +76,9 @@ def clean(args):
     staged_file_contents = OrderedDict({"model_dir": theta_model_dir})
     # Sort the keys so we don't get changing diffs based on serialization order.
     for keys, param in sorted(utils.flatten(model_checkpoint).items()):
-        # TODO(bdlester): We should update this staged_file_content from having
-        # different keys for shape, dtype, and hash to a dict.
-        param_name = "/".join(keys)
-        staged_file_contents[f"{param_name} shape"] = params.get_shape_str(param)
-        staged_file_contents[f"{param_name} dtype"] = params.get_dtype_str(param)
-        staged_file_contents[f"{param_name} hash"] = params.get_hash(param)
+         param_name = "/".join(keys)
```
blester125, line 79: It looks like you have 5 spaces here instead of 4.

We need to get black running (see https://github.com/r-three/git-theta/issues/88) but until then can you fix this formatting?

muqeeth: Thanks for the catch. I changed it

bin/git-theta-filter
```diff
@@ -76,9 +76,9 @@ def clean(args):
     staged_file_contents = OrderedDict({"model_dir": theta_model_dir})
     # Sort the keys so we don't get changing diffs based on serialization order.
     for keys, param in sorted(utils.flatten(model_checkpoint).items()):
-         param_name = "/".join(keys)
-         tensor_metadata = OrderedDict({"shape": params.get_shape_str(param), "dtype": params.get_dtype_str(param), "hash": params.get_hash(param)})
-         staged_file_contents[param_name] = OrderedDict({"tensor_metadata": tensor_metadata}) 
+        param_name = "/".join(keys)
+        tensor_metadata = OrderedDict({"shape": params.get_shape_str(param), "dtype": params.get_dtype_str(param), "hash": params.get_hash(param)})
```
blester125, line 80: This ordered dict isn't actually ensuring the order will be consistent.

The ordered dict records and will iterate based on insertion order, but the normal dict (which is created because arguments are evaluated before being passed) you are making as a parameter to the ordered dict init does not (pre python3.7). These when the orderdict iterates through the passed dict it will record that particular order which could be different between runs.

I think you need to manually make calls like `...["shape"] = ...` to ensure it will always be the same order.

blester125 (approved):
LGTM

PR #92, opened 2022-11-23
Add support for multiple update types via plugins

blester125:
This PR adds support for parameter updates, where the parameter value at timestep t is computed as an update to timestep t-1. The update value is what is stored on disk instead of the actual parameter at timestep t. When the parameter at timestep t in reconstituted in a git smudge, it is computed by applying the update (loaded from disk) to the parameter at timestep t-1.

We track updates on the filesystem using a slightly updated structure. Each network parameter still has a `.../params/` directory, but instead of the parameters living directly in that directory, they live in a `.../params/updates/${hash}/` directory. This hash is the hash of the parameters at timestep t, not the hash of the update on disk.

Additionally there is now a `.../params/metadata` file that tracks what the last update to be made to these parameters was. There is also a `.../params/updates/${hash}/metadata` file. This tracks what kind of update was used and what the previous update was.

Now reconstituting via a smudge is done by creating the update plugin used for the most recent update and calling `.apply` on it. This will get the value of the parameter after that update was applied. True updates, like sparse, will recursively get the parameter value at timestep t-1 by creating and applying the update plugin used in the previous update. It will then apply it's own update to the result.

Updates that touch all the parameters, for example a dense update, act as base cases. They don't track what the last update was or call plugin used for the last update, they just return the parameter values as they were at some timestep i (read directly from disk).

I opted to implement this recursive update (where each update keeps a backpointer to the update before it) instead of a list of updates implementation we has discussed as it allow us to skip the creation of parameter t-1 when the update at t is not a true update (dense).

The on-disk format for parameters is more complex/verbose now, but the shared format for different update types makes manipulation easy, additionally, naming updates by the hash of the resulting parameter (instead of the update values) should allow for an easy implementation for things like `git theta compact` which could collapse a series of sparse updates into a single update.

This PR also adds some tests to the cascading override behavior we use when determining what update type to use based on various methods of capturing user input.

### DIFF
--- a/git_theta/updates/dense.py
+++ b/git_theta/updates/dense.py
@@ -56,6 +56,8 @@ class DenseUpdate(Update):
             The values to write.
         """
         # Remove any past updates to begin anew.
+        # TODO: Move to using git_utils.remove_file once the edge case of the
+        # path not existing is git is resolved.
         shutil.rmtree(path)
         parameter_hash = params.get_hash(parameter)
         output_dir = os.path.join(path, UpdateConstants.UPDATES_DIR, parameter_hash)
--- a/git_theta/updates/dense.py
+++ b/git_theta/updates/dense.py
@@ -1,0 +1,92 @@
+"""Class managing dense parameter updates."""
+
+import logging
+import os
+import shutil
+from typing import Optional
+import numpy as np
+from git_theta import file_io
+from git_theta import params
+from git_theta.updates import Update, UpdateConstants
+
+
+class DenseUpdate(Update):
+    """An update where all parameters are changed."""
+
+    @property
+    def name(self):
+        """The name used to get this class as a plugin."""
+        return "dense"
+
+    def read(self, path: str) -> np.ndarray:
+        """Read the parameter values in the path dir.
+
+        Note:
+            In this case, the read parameter values are the whole update, all
+            values are overwritten so the result of this read will just be
+            returned.
+
+        Parameters
+        ----------
+        path
+            The .../params/updates/${hash} directory path that contains parameter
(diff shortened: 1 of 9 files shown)


git_theta/updates/dense.py
```diff
@@ -0,0 +1,92 @@
...
+
+        Parameters
+        ----------
+        path
+            The .../params/updates/${hash} directory path to write to.
+        parameter
+            The values to write.
+        """
+        # Remove any past updates to begin anew.
+        shutil.rmtree(path)
```
nkandpa2, line 59: Should this be `git_utils.remove_file`? Not sure if git will stage the previous contents of `updates/` to be removed after `shutil.rmtree`.

blester125: The root dir of all this parameter stuff (`path` here, i.e. `.../param/`) is eventually the target of a git add. That add on the dir will stage the deletion of all the contents of `.../param/updates` as well as the addition of the new values.

Can the remove file function handle directories?

nkandpa2: That makes sense. `remove_file` (bad name) will run `git rm -r` if the argument is a directory. Right now the way it's called in `git theta add` it gets called on the `params/` directory.

nkandpa2 (approved):

git_theta/updates/dense.py
blester125: Currently using `remove_file` doesn't handle the case where there is no tracked file well (see https://github.com/r-three/git-theta/issues/104) so I added a TODO to switch to it from shutil once it handles that edges case

Issue #91, opened 2022-11-23
Add `.name` as a `@property` to our checkpoint handler objects


blester125:
We should record a `.name` value on each of the plugins. This property on a checkpoint object should return a string so that a `get_checkpoint` function call with this name would return the class of this object.

This will make things like logging what checkpoint type is used (and making sure we use the same one across multiple cleans, etc) much easier, especially when the value is set via an environment variable.

PR #98, opened 2022-11-23
Adds a `.name` property to checkpoint handlers.

blester125:
This PR adds a `.name` property to checkpoint handlers. This name can be used to look up the checkpoint handler as a plugin. This property lets us know what handler is being used (for example in logging) even when the handler selection was done with cascading override (for example the environment variable was used instead of cli based user input). i.e. Adding this lets us know that a pickled dict checkpoint handler was used, even if the actual user input was `None`.

closes https://github.com/r-three/git-theta/issues/91


bin/git-theta
```diff
@@ -52,13 +52,15 @@ def add(args):
...
     """
+    logging.debug(f"Running git theta add on {args.file}")
     repo = git_utils.get_git_repo()
     model_path = git_utils.get_relative_path_from_root(repo, args.file)
     theta_model_dir = git_utils.get_git_theta_model_dir(repo, model_path, create=True)
 
     # TODO: Don't default to pytorch once other checkpoint formats are supported.
     checkpoint_type = args.type or "pytorch"
     checkpoint = checkpoints.get_checkpoint(checkpoint_type)
+    logging.debug(f"git theta add using checkpoint type'={checkpoint.name}'")
```
nkandpa2, line 63: Extra single-quotes in log message?

nkandpa2 (approved):

bin/git-theta
blester125: fixed

Issue #107, opened 2022-11-30
Update `Update` plugin classes


blester125:
As have talked about what more update types will look like and it seems the update classes can be updated to codify this a bit more.

Update classes should have a `calculate_difference` method.

Things like `apply` and maybe `write` in the base class should probably be updated to have the actual code with calls to abstract methods the subclasses implement. Instead of each update type implementing all of `apply` on their own.

blester125:
This was done as part of #121

PR #112, opened 2022-12-04
Refactor Update plugins

blester125:
This PR updates the Update plugin base classes. There is still the original base-class with overrideable read, write, and apply, but there is also a new TrueUpdate class which has definitions for read, write, and apply that work under the assumption that this update is handled using the previous value. This class includes a few smaller method about calculating and applying updates to increase code reuse between true update plugins.

The new implementation allow for writing new checkpoint types really easily and makes understanding the implementation of the true updates easier. I looked at trying to include this in the original Base class but some of the Dense updates implementations of the helper methods would be strange NoOps and not really make sense.

This PR also updates the format of the parameter level metadata file (it is now a dict) to allow unification of "get most recent update" when talking about the whole parameter or the most recent update before the current update.

### DIFF
--- a/git_theta/updates/__init__.py
+++ b/git_theta/updates/__init__.py
@@ -1,6 +1,7 @@
 """Classes for controlling how parameter updates are made."""
 
 from git_theta.updates.base import (
+    TrueUpdate,
     Update,
     get_update_handler,
     UpdateConstants,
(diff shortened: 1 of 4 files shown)


blester125:
Note: Tests should pass after https://github.com/r-three/git-theta/pull/116 is merged.

blester125:
closes https://github.com/r-three/git-theta/issues/107

blester125:
Changes to how updates work in https://github.com/r-three/git-theta/pull/114 make this out of date, main ideas rolled into https://github.com/r-three/git-theta/pull/121

Issue #111, opened 2022-12-02
Proposal for fixing issues related to coupling between model checkpoints and `.git_theta/`


nkandpa2:
# Problems
The current design couples the state of a tracked checkpoint on disk with the contents of `.git_theta/` and tracks the contents of most of `.git_theta` using git LFS. I feel like lately we've been running into issues that arise from this coupling of checkpoints and `.git_theta`. Here are a few of the issues that we've hit and general inefficiencies of doing things this way:

1. When we smudge a checkpoint, we rely on `.git_theta/` having the correct contents. This happens most of the time since git populates the working tree in lexical order and model checkpoints usually have names that come after `.git_theta` lexically. However, this is an undocumented feature that should not be relied upon.
2. When we want to diff or merge two checkpoints, their representations in `.git_theta/` are useful since they give information about what updates were applied between the two checkpoints. However, at the time that the diff and merge tools run, only one of the versions (the currently checked out version) can be in `.git_theta/`. 
3. When we stage a model that is $n$ bytes on disk, the effective size of the repo is about $3n$ bytes which seems unnecessary (one copy in the working tree, one copy in `.git_theta/`, and one copy in `.git/lfs` since the contents of `.git_theta/` are tracked by LFS)
4. Probably some issues in the future

The core of some of many of these issues is that git is not meant to be used in this way where each version of a file is tied to a corresponding version of other files, since git treats each file independently for all of its operations (checkout, diff, merge, etc.). Here is a proposal to fix these issues.

# Implementation
#### Upon `git add <checkpoint>` the clean filter should:
1. Load the checkpoint into memory
2. Write each parameter group to a temporary file
3. Pass each temporary file to `git lfs clean`. This command will copy the contents of the file to `.git/lfs/objects` and output the contents of an LFS pointer file to stdout.
4. Save the output of `git lfs clean` (i.e., the contents of the pointer files). This output contains the git LFS spec version, the file's LFS oid (object id), and the file size.
5. Create a metadata file of the following form:
```json
{
    "param_group1": {
        "oid": "object id 1",
        "size": 1000
    }
    "param_group2": {
        "oid": "object id 2",
        "size": 500
    }
}
```
Basically each parameter group name maps to a dictionary containing the information that git LFS would normally put in that parameter group's pointer file.

#### Upon `git push` the pre-push hook should:
1. Iterate over all files being committed
2. Find the git-theta tracked files (i.e., see if they match the git-theta metadata file format)
[...]

PR #114, opened 2022-12-05
Multi pointer

nkandpa2:
Implementation of proposal described in #111


git_theta/file_io.py
```diff
@@ -2,11 +2,24 @@
...
 import json
 import logging
+import zipfile
+import tarfile
 
 from file_or_name import file_or_name
 
 
-def load_tracked_file_from_memory(files):
+def untar_tracked_file_in_memory(file_bytes):
```
blester125, line 11: What in the goal of taring and untaring? I think different goals might lead to different choices of compression libraries

* To make a single object git-lfs stores
* To reduce the disk usage
* To reduce I/O bottlenecks (less data over pipes/to files)

bin/git-theta
```diff
@@ -43,45 +51,53 @@ def parse_args():
...
-
-    for param_file in utils.removed_params(param_dict, utils.walk_parameter_dir(theta_model_dir)):
-        git_utils.remove_file(param_file, repo)
+    gitattributes_file = git_utils.get_gitattributes_file(repo)
+    patterns = git_utils.get_gitattributes_tracked_patterns(gitattributes_file)
 
-    git_utils.add_file(model_path, repo)
+    # Read lines of the form <local ref> <local sha1> <remote ref> <remote sha1> LF
+    lines = sys.stdin.readlines()
+    lines_parsed = [re.match('^(?P<local_ref>[^\s]+) (?P<local_sha1>[^\s]+) (?P<remote_ref>[^\s]+) (?P<remote_sha1>[^\s]+)\n$', l) for l in lines]
```
blester125, line 88: We probably want a tighter regex than space splitting. Something like

`\b[0-9a-f]{40}\b`

I assume this format is something we don't control? like it is from git so we can't convert it into json or anything?

bin/git-theta
```diff
@@ -110,4 +122,5 @@ def track(args):
 
 if __name__ == "__main__":
     args = parse_args()
+    git_utils.set_hooks()
```
blester125, line 125: Should setting hooks be part of the install command? Or is there a reason to run it everytime?

git_theta/git_utils.py
```diff
@@ -193,74 +195,66 @@ def remove_file(f, repo):
         repo.git.rm(f)
 
 
-def git_lfs_install():
-    """
-    Run the `git lfs install` command
+def get_file_version(repo, path, commit_hash):
+    try:
+        commit = repo.commit(commit_hash)
```
blester125, line 200: I think you can do item access on the tree objects, like `repo.commit(commit_hash).tree[path]`

git_theta/hooks/post-commit
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
```
blester125, line 1: I think you need a `__init__.py` file for the `__path__` you use later to work in a non-editable install.

bin/git-theta
nkandpa2: Yeah, nice catch. It shouldn't be run as part of install since there isn't any repo to put the hooks into.

bin/git-theta
nkandpa2: Yeah, this can be more precise. The format is from the githooks documentation: https://mirrors.edge.kernel.org/pub/software/scm/git/docs/githooks.html

<local ref> SP <local object name> SP <remote ref> SP <remote object name> LF

Object names we know the length of so we can enforce that.

git_theta/hooks/post-commit
nkandpa2: Just an empty `__init__.py`, right? Nothing can be imported since the files are just shell scripts. 

Aside from that, is this `__path__` method the best way of doing this?

blester125: It looks like the current recommended way to do this is

1) Update the setup.py to include these files in the `package_data` dict
2) Include these files in a MANIFEST.in
3) iterate through importlib.resources.files to find the files https://docs.python.org/3/library/importlib.resources.html#importlib.resources.files
4) Treat them as files with importlib.resoures.as_file https://docs.python.org/3/library/importlib.resources.html#importlib.resources.as_file

So I think this would end up doing a read+write instead of just a copy.

The above is the most general way to do it. I think you way also works if we set `zip_safe` to false in the setup (but then it is weird that you have this hooks import when there isn't really anything to import)

bin/git-theta
```diff
@@ -108,6 +124,19 @@ def track(args):
     git_utils.add_file(gitattributes_file, repo)
 
 
+def add(args, unparsed_args):
```
blester125, line 127: What is the reason for manually using subprocess.run to talk to git over the GitPython `repo.git.add` command? 

It looks like already has an environment variable context manager, from the docs:

```python
ssh_cmd = 'ssh -i id_deployment_key'
with repo.git.custom_environment(GIT_SSH_COMMAND=ssh_cmd):
    repo.remotes.origin.fetch()
```

git_theta/file_io.py
nkandpa2: Mostly 1 and 3. Conceptually feels a bit simpler if one parameter group corresponds to one lfs object. Also fewer calls to the lfs smudge filter. The disk usage isn't affected since tar is being used without compression.

bin/git-theta
nkandpa2: Good call, simpler to just use gitpython.

git_theta/hooks/post-commit
nkandpa2: Did some testing and it doesn't seem like adding MANIFEST.in was necessary. Adding the scripts to `package_data` resulted in the hooks being put into the installed package. Also, `importlib.resources.as_file` gives back a path so we can still just copy with `shutil`.

I am fairly inexperienced with packaging in python so I may have missed something here. The way I was testing was just running `pip install .` from the git-theta repo and then verifying that commands that copy the hooks into place were actually doing this.

bin/git-theta-filter
```diff
@@ -60,26 +53,43 @@ def clean(args):
...
-    # Sort the keys so we don't get changing diffs based on serialization order.
-    for keys, param in sorted(utils.flatten(model_checkpoint).items()):
-        param_name = "/".join(keys)
-        tensor_metadata = OrderedDict({"shape": params.get_shape_str(param)})
-        tensor_metadata["dtype"] = params.get_dtype_str(param)
-        tensor_metadata["hash"] = params.get_hash(param)
-        staged_file_contents[param_name] = OrderedDict({"tensor_metadata": tensor_metadata})
+    obj = git_utils.get_file_version(repo, args.file, "HEAD")
+    if obj:
+        curr_metadata = metadata.Metadata.from_file(obj.data_stream).flatten()
```
blester125, line 63: `prev_metadata` is probably a clearer name?

bin/git-theta
```diff
@@ -39,49 +44,64 @@ def parse_args():
...
+def post_commit(args):
+    repo = git_utils.get_git_repo()
+    theta_commits = theta.ThetaCommits(repo)
+
+    gitattributes_file = git_utils.get_gitattributes_file(repo)
+    patterns = git_utils.get_gitattributes_tracked_patterns(gitattributes_file)
+
+    oids = set()
+    commit = repo.commit('HEAD')
+    for obj in commit.tree.traverse():
```
blester125, line 66: This seems to iterate over all the objects in the git repo. Is there an issue with using `commit.stats.files` to just check the files that were changed in this commit?

bin/git-theta
```diff
@@ -39,49 +44,64 @@ def parse_args():
...
+        if any([fnmatch.fnmatchcase(obj.path, pattern) for pattern in patterns]):
+            curr_metadata = metadata.Metadata.from_file(obj.data_stream).flatten()
+
+            prev_obj = git_utils.get_file_version(repo, obj.path, 'HEAD~1')
```
blester125, line 70: Does this break with `git mv`? In that case the path at `"HEAD~1"` might be different so `prev_obj` would result in being `None`?

There is a diff index object you get with `diff_index = commit1.diff(commit2)` and the diff objects in it have things like `a_path` and `b_path` which might help us fix this edge case (if it is broken)

bin/git-theta
```diff
@@ -39,49 +44,64 @@ def parse_args():
...
+            if prev_obj:
```
blester125, line 71: We seem to have this check in a few places. We could add a `from_commit` class method to the metadata file that handles getting a file version, checking if it is none, etc.

bin/git-theta
```diff
@@ -39,49 +44,64 @@ def parse_args():
...
+    add_parser.set_defaults(func=add)
+
+    args = parser.parse_known_args()
     return args
 
 
-def paths_and_parameters(model_dict, root):
-    """Convert the keys of the model dict to parameters based starting at root."""
-    return {os.path.join(root, *k): v for k, v in utils.flatten(model_dict).items()}
+def post_commit(args):
```
blester125, line 57: Probably add a doc string for this

bin/git-theta
```diff
@@ -39,49 +44,64 @@ def parse_args():
...
+    theta_commits = theta.ThetaCommits(repo)
 
-    for param_file in utils.removed_params(param_dict, utils.walk_parameter_dir(theta_model_dir)):
-        git_utils.remove_file(param_file, repo)
-
-    git_utils.add_file(model_path, repo)
+    # Read lines of the form <local ref> <local sha1> <remote ref> <remote sha1> LF
+    lines = sys.stdin.readlines()
+    print(lines)
+    lines_parsed = [re.match('^(?P<local_ref>[^\s]+)\s+(?P<local_sha1>[a-f0-9]{40})\s+(?P<remote_ref>[^\s]+)\s+(?P<remote_sha1>[a-f0-9]{40})\s+$', l) for l in lines]
```
blester125, line 92: We should abstract this parsing into a function that we can test.

bin/git-theta
```diff
@@ -108,6 +124,18 @@ def track(args):
     git_utils.add_file(gitattributes_file, repo)
 
 
+def add(args, unparsed_args):
```
blester125, line 127: What is the point of still having this `git theta add` command? Is it the only way to include user input like update type?

Does this mean to do a sparse update you need to do `git theta add` but `git add` will always assume a dense update?

git_theta/updates/base.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def calculate_update(self, repo, path, param_keys, param_metadata, param):
+        raise NotImplementedError
+
+
+def get_update_name(update_type: Optional[str] = None) -> str:
+    return update_type or os.environ.get(utils.EnvVarConstants.UPDATE_TYPE) or "dense"
+
+
+def get_update(update_type: Optional[str] = None) -> Update:
```
blester125, line 52: You should rebase on main or copy some of the changes over, for example, this has been renamed `get_update_handler` to be more in-line with the checkpoint one.

git_theta/updates/base.py
```diff
@@ -0,0 +1,68 @@
...
+else:
+    from importlib.metadata import entry_points
+
+import logging
+from typing import Optional
+
+from git_theta import git_utils, utils, params, metadata
+
+
+class Update:
```
blester125, line 17: I have some ideas on how to refactor the Update class to be more encapsulated (and it should help making things async eventually) but those are big changes. We should punt until this is landed.

blester125 (changes requested):
Added some naming suggestions and documentation requests.

The branch is also pretty far behind head, it should be pulled to include things like new names, etc

git_theta/metadata.py
```diff
@@ -0,0 +1,167 @@
...
+        )
+        return cls(
+            version=match.group("version"),
+            oid=match.group("oid"),
+            size=match.group("size"),
+        )
+
+    @classmethod
+    def from_bytes(cls, b):
+        return cls.from_pointer(git_utils.git_lfs_clean(b))
```
blester125, line 52: `git_lfs_clean` is the code that actually takes the `b` bytes and sends them to git lfs right? And that results in the bytes getting put into the lfs store?

I think it is surprising for a big chunk of computation/communication to happen inside the creation method for the metadata object. It seems like it should be a cheap operation and it also makes it hard to tell where the writing happens. I think this is fine for now but we may want to change it if we refactor the updates a bit to make the git-lfs write more explicit.

git_theta/metadata.py
```diff
@@ -0,0 +1,167 @@
...
+@dataclasses.dataclass(eq=True)
+class TensorMetadata(MetadataField):
+    shape: str
+    dtype: str
+    hash: str
+
+    @classmethod
+    @property
+    def name(cls):
+        return "tensor_metadata"
```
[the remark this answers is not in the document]
blester125: NVM, I see that the name is used as a key in the metadata file, it makes sense to have `_metadata` in that case.

git_theta/metadata.py
```diff
@@ -0,0 +1,167 @@
...
+
+    @classmethod
+    def from_tensor(cls, tensor):
+        shape = str(tensor.shape)
+        dtype = str(tensor.dtype)
```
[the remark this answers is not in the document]
blester125: `str(a.dtype)` may be better IIRC `<` is an indication of endiness, that is important for tensorstore but not for our tensor comparisons.

git_theta/git_utils.py
```diff
@@ -193,74 +196,45 @@ def remove_file(f, repo):
...
-    if not pattern_found:
-        new_gitattributes.append(f"{path} filter=theta")
-    return new_gitattributes
+def git_lfs_push_oids(remote_name, oids):
+    if len(oids):
+        out = subprocess.run(
+            ["git", "lfs", "push", "--object-id", remote_name] + list(oids)
+        )
+        return out.returncode
+    else:
```
blester125, line 239: We don't need this `else:`

git_theta/git_utils.py
```diff
@@ -193,74 +196,45 @@ def remove_file(f, repo):
...
-    """
-    out = subprocess.run(["git", "lfs", "install"])
-    return out.returncode
 
+def get_head(repo):
+    try:
+        head = repo.commit("HEAD")
+        return head.hexsha
+    except git.BadName:
+        None
```
blester125, line 216: Probably do an explicit `return None` instead of just `None`, in this case `None` is returned but that is just because it is the default value if there are no returns.

git_theta/utils.py
```diff
@@ -1,17 +1,20 @@
...
 
 class EnvVarConstants:
     CHECKPOINT_TYPE: str = "GIT_THETA_CHECKPOINT_TYPE"
     UPDATE_TYPE: str = "GIT_THETA_UPDATE_TYPE"
 
 
-def flatten(d: Dict[str, Any]) -> Dict[Tuple[str, ...], Any]:
+def flatten(
+    d: Dict[str, Any],
+    is_leaf: Callable[[object], bool] = lambda v: not isinstance(v, dict),
```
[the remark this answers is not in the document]
nkandpa2: Fixed

git_theta/utils.py
```diff
@@ -1,17 +1,20 @@
 """Utilities for git theta"""
 
 
-import operator as op
 import os
 from typing import Dict, Any, Tuple, Union, Callable
+import contextlib
```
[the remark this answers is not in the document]
nkandpa2: Fixed

git_theta/theta.py
```diff
@@ -0,0 +1,82 @@
+import os
```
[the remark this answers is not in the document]
nkandpa2: Added

setup.py
```diff
@@ -45,7 +45,8 @@ def get_version(file_name: str, version_variable: str = "__version__") -> str:
     author="Colin Raffel",
     author_email="craffel@gmail.com",
     url="https://github.com/r-three/checkpoint-vcs",
-    packages=["git_theta"],
+    packages=["git_theta", "git_theta.updates"],
```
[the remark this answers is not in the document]
nkandpa2: method got removed in refactor

git_theta/theta.py
```diff
@@ -0,0 +1,82 @@
...
+    def load_commit_file(f):
+        return json.load(f)
+
+    @staticmethod
+    @file_or_name(f="w")
+    def write_commit_file(f, oids):
+        return json.dump(list(oids), f)
+
+    @staticmethod
+    def get_oids(commit):
```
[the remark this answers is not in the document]
nkandpa2: Agreed, the naming isn't super clear and consistent with the things being returned. I refactored this module as follows:
1. The data being written to the .git/theta/commits/* files have been factored out into its own CommitInfo class to make it simpler to modify what's being written if we want to store more commit metadata in these files in the future.
2. All get_commit_info* methods take commit hashes and return CommitInfo objects
3. All get_commit_oid* methods take commit hashes and return sets of oids

git_theta/theta.py
```diff
@@ -0,0 +1,82 @@
...
+    def get_commit(self, commit_hash):
+        path = self.get_commit_path(commit_hash)
+        if not (os.path.exists(path) and os.path.isfile(path)):
+            raise ValueError(f"commit {commit_hash} is not found in {self.path}")
+        commit = self.load_commit_file(path)
+        return commit
+
+    def get_commit_range(self, start_commit, end_commit):
+        logging.debug(f"Getting commits from {start_commit}..{end_commit}")
+        if re.match("^0+$", start_commit):
```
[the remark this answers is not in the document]
nkandpa2: Added a comment to explain:

`# N.b. the all-zero hash is used by git to indicate a non-existent start hash`
`# For example, a git pre-push hook will receive the all-zero hash if the remote ref does not have any commit history`

git_theta/theta.py
```diff
@@ -0,0 +1,82 @@
...
+        return os.path.join(self.path, commit_hash)
+
+    def get_commit(self, commit_hash):
+        path = self.get_commit_path(commit_hash)
+        if not (os.path.exists(path) and os.path.isfile(path)):
+            raise ValueError(f"commit {commit_hash} is not found in {self.path}")
+        commit = self.load_commit_file(path)
+        return commit
+
+    def get_commit_range(self, start_commit, end_commit):
```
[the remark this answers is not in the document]
nkandpa2: Refactored and used more consistent names with the return type

git_theta/theta.py
```diff
@@ -0,0 +1,82 @@
...
+        if re.match("^0+$", start_commit):
+            commits = self.repo.iter_commits(end_commit)
+        else:
+            commits = self.repo.iter_commits(f"{start_commit}..{end_commit}")
+
+        commits = [commit.hexsha for commit in commits]
+        logging.debug(f"Found commits {commits}")
+        return commits
+
+    def get_commit_oids(self, commit_hash):
```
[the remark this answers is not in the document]
nkandpa2: Honestly not sure what it should be. Right now the object is used in two ways:

1. The caller uses `get_commit_oids_ranges` to get a set of oids to push to to the remote with `git_utils.git_lfs_push`. So the `get_commit_oids*` methods should probably be exposed.
2. The caller creates a `CommitInfo` object and then calls `write_commit_info` to write that object to .git/theta/commits.

Right now we don't have any need to expose the `get_commit_info*` methods but we may need to in the future depending on whether we extend the types of metadata being stored in .git/theta/commits and need to access that from some function.

git_theta/params.py
```diff
@@ -1,46 +1,106 @@
...
-        hash of parameter bytes
-    """
-    return hashlib.sha1(np.asarray(p).tobytes()).hexdigest()
+"""Classes for serializing model updates."""
+
+from collections import defaultdict
+import tensorstore as ts
+import io
+import tarfile
+import posixpath
```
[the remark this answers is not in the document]
nkandpa2: I believe tar can handle windows style paths as well. Posixpath is explicitly used here since I wanted to make sure the virtual paths in the tar file are consistent when people read and write them from both unix and windows. If the code used os.path instead then windows users wouldn't be able to properly split the virtual paths created by unix users and vice-versa.

git_theta/params.py
```diff
@@ -1,46 +1,106 @@
...
+                    archive.addfile(tarinfo, io.BytesIO(file_bytes))
+
+        tarred_file.seek(0)
+        return tarred_file.read()
+
+    def split(self, tarred_file):
+        file = io.BytesIO(tarred_file)
+        param_files = defaultdict(dict)
+        with tarfile.open(fileobj=file, mode="r") as archive:
+            for archive_file in archive.getnames():
```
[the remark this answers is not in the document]
nkandpa2: Fixed

git_theta/metadata.py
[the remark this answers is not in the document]
nkandpa2: Agreed, I think `str(dtype)` is preferred

git_theta/metadata.py
```diff
@@ -0,0 +1,167 @@
...
+import dataclasses
+from collections import OrderedDict
+import re
+import json
+
+from git_theta import git_utils, utils
+from file_or_name import file_or_name
+
+
+class MetadataField:
```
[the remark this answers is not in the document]
nkandpa2: Fixed

git_theta/metadata.py
```diff
@@ -0,0 +1,167 @@
...
+    def to_serializable(self):
```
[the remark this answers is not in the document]
nkandpa2: Agreed. Fixed.

git_theta/metadata.py
```diff
@@ -0,0 +1,167 @@
...
+
+    @property
+    def lfs_pointer(self):
+        return (
+            f"version {self.version}\noid sha256:{self.oid}\nsize {self.size}".encode()
+        )
+
+    @classmethod
+    def from_pointer(cls, pointer_file):
+        match = re.match(
```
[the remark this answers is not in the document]
nkandpa2: Fixed

git_theta/metadata.py
```diff
@@ -0,0 +1,167 @@
...
+        return "lfs_metadata"
+
+    @property
+    def lfs_pointer(self):
+        return (
+            f"version {self.version}\noid sha256:{self.oid}\nsize {self.size}".encode()
+        )
+
+    @classmethod
+    def from_pointer(cls, pointer_file):
```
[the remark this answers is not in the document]
nkandpa2: Fixed variable name

git_theta/metadata.py
```diff
@@ -0,0 +1,167 @@
...
+        match = re.match(
+            "^version (?P<version>[^\s]*)\s*oid sha256:(?P<oid>[^\s]*)\s*size (?P<size>[0-9]*)$",
```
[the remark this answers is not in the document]
nkandpa2: Took a closer look at https://github.com/git-lfs/git-lfs/blob/main/docs/spec.md that lays out the rules for pointer files. Restricted the regex to only match `\n`, added the trailing `\n`, and made sha256 pattern more strict. All three keys are required and required in that order. There could be additional keys, but not sure whether we should be handling that with a regex.

git_theta/metadata.py
```diff
@@ -0,0 +1,167 @@
...
+
+    @classmethod
+    @property
+    def name(cls):
+        return "lfs_metadata"
+
+    @property
+    def lfs_pointer(self):
+        return (
+            f"version {self.version}\noid sha256:{self.oid}\nsize {self.size}".encode()
```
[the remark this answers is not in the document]
nkandpa2: Fixed

git_theta/metadata.py
```diff
@@ -0,0 +1,167 @@
...
+import json
+
+from git_theta import git_utils, utils
+from file_or_name import file_or_name
+
+
+class MetadataField:
+    def to_serializable(self):
+        d = dataclasses.asdict(self)
+        for k, v in d.items():
```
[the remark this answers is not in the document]
nkandpa2: Fixed

git_theta/metadata.py
```diff
@@ -0,0 +1,167 @@
...
+        return cls(tensor_metadata, lfs_metadata, theta_metadata)
+
+    def to_serializable(self):
+        for metadata_key, metadata in self.items():
+            self[metadata_key] = metadata.to_serializable()
+        return self
+
+    @property
+    def theta_metadata(self):
+        return self.get(ThetaMetadata.name)
```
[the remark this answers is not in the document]
nkandpa2: Fixed

git_theta/metadata.py
```diff
@@ -0,0 +1,167 @@
...
+    update_type: str
+    last_commit: str
+
+    @classmethod
+    @property
+    def name(cls):
+        return "theta_metadata"
+
+
+class ParamMetadata(OrderedDict):
```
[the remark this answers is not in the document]
nkandpa2: Made this a data class which did simplify serialize in the base class.

git_theta/metadata.py
```diff
@@ -0,0 +1,167 @@
...
+        for param_keys in set(self_flattened.keys()).intersection(
+            other_flattened.keys()
+        ):
+            if (
+                self_flattened[param_keys].tensor_metadata
+                != other_flattened[param_keys].tensor_metadata
+            ):
+                modified.append(self_flattened[param_keys])
+
+        return added, removed, modified
```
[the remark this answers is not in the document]
nkandpa2: Agreed. Changed it to return three `Metadata` objects with only the fields that were added/removed/modified. Callers can inspect those to get oids or whatever else they care about.

git_theta/metadata.py
```diff
@@ -0,0 +1,167 @@
...
+            size=match.group("size"),
+        )
+
+    @classmethod
+    def from_bytes(cls, b):
+        return cls.from_pointer(git_utils.git_lfs_clean(b))
+
+
+@dataclasses.dataclass(eq=True)
+class TensorMetadata(MetadataField):
```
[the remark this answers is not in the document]
nkandpa2: Agreed. Since there's going to be a good bit of testing migration and testing of new functions, let's make that its own issue.

git_theta/metadata.py
```diff
@@ -0,0 +1,167 @@
+"""Classes representing checkpoint metadata files"""
+
+import hashlib
+import dataclasses
```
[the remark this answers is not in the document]
nkandpa2: Installed backport in `setup.py`

git_theta/metadata.py
```diff
@@ -0,0 +1,167 @@
...
+    @property
+    def tensor_metadata(self):
+        return self.get(TensorMetadata.name)
+
+    @property
+    def lfs_metadata(self):
+        return self.get(LfsMetadata.name)
+
+
+class Metadata(OrderedDict):
```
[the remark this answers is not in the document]
nkandpa2: This one can't be a data class since it doesn't have a fixed set of keys (i.e. each checkpoint will have a different set of parameters)

git_theta/git_utils.py
```diff
@@ -193,74 +196,45 @@ def remove_file(f, repo):
...
-        new_gitattributes.append(line)
-    # If we don't find a matching pattern, add a new line that covers just this
-    # specific file.
-    if not pattern_found:
-        new_gitattributes.append(f"{path} filter=theta")
-    return new_gitattributes
+def git_lfs_push_oids(remote_name, oids):
+    if len(oids):
+        out = subprocess.run(
+            ["git", "lfs", "push", "--object-id", remote_name] + list(oids)
```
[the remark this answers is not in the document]
nkandpa2: The pre-push hook gets called with two arguments, a remote name and a remote location. Remote name is something like `origin`, whereas remote location would be the URL. If the remote does not have a name, then both arguments will be the same. So the argument to this function could either be a name like `origin` or a URL. I guess we'd need to escape this if any `\` characters appear. Do those typically appear in remote URLs?

git_theta/git_utils.py
```diff
@@ -26,58 +30,13 @@ def get_git_repo():
...
-        os.makedirs(git_theta_model_dir)
-
-    return git_theta_model_dir
+def set_hooks():
+    repo = get_git_repo()
+    hooks_dst = os.path.join(repo.git_dir, "hooks")
+    package = importlib.resources.files("git_theta")
+    for hook in ["pre-push", "post-commit"]:
+        with importlib.resources.as_file(package.joinpath("hooks", hook)) as file:
+            shutil.copy(file, hooks_dst)
```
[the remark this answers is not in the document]
nkandpa2: Yeah this implementation will just replace any existing hooks (either a user's hooks or the git-theta hooks put there by a previous command) any time a git-theta command (except install) is run. I'll add a check in to only copy if the hook scripts are not already there. I think for the second problem (overwriting existing user hooks), we should take an approach similar to this https://stackoverflow.com/questions/26624368/handle-multiple-pre-commit-hooks but since this is a bit more involved let's make this it's own issue.

git_theta/git_utils.py
```diff
@@ -3,13 +3,17 @@
...
 import git
 import os
+import stat
 import json
 import logging
 import io
 import re
 import torch
 from typing import List, Union
+from collections import OrderedDict
```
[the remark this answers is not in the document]
nkandpa2: Fixed

git_theta/git_utils.py
```diff
@@ -3,13 +3,17 @@
 import fnmatch
 import git
 import os
+import stat
```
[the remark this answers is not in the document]
nkandpa2: Fixed

bin/git-theta-filter
nkandpa2: Fixed

git_theta/git_utils.py
nkandpa2: Lol whoops, fixed

git_theta/metadata.py
nkandpa2: Agreed. One possibility is to use the `git lfs pointer` command that only constructs the pointer file without copying to `.git/lfs`. The command gives a little extra output that would need to be stripped away before passing to `from_pointer`.

bin/git-theta
nkandpa2: Good idea. Fixed

bin/git-theta
nkandpa2: Good idea. Fixed.

bin/git-theta
nkandpa2: The intention was for this to be a convenience command for specifying update type. Right now we're giving this information to git-theta through environment variables but it seems more convenient to run `git theta add model.pt --update-type sparse` than `GIT_THETA_UPDATE_TYPE=sparse; git add model.pt`. The default behavior for when the environment variable isn't set is just dense.

bin/git-theta
nkandpa2: This is a great point. A couple things definitely break with `git mv`. 

1. post-commit hook uses the metadata file path to figure out what LFS oids were changed in the previous commit (NOTE: this requires looking at the parent or parents of the HEAD commit and inspecting the metadata files)
2. the clean filter uses the path to the metadata file to look at the state of the checkpoint at HEAD
3. the `Update` class uses the path to the metadata file to get the state at a previous commit in `Update.get_last_version`.

Basically, anywhere that we look for a version of the metadata file, we implicitly rely on the fact that the path to it has not changed. We should write up a unified solution for all three places we do this that uses the diff index object you'd mentioned. Since this is a bit tricky, let's push this to another issue.

bin/git-theta
nkandpa2: Done.

bin/git-theta
nkandpa2: Done.

git_theta/updates/base.py
nkandpa2: Manually copied/checked-out the stuff that's been incorporated into main since this branch was started.

blester125 (approved):

Issue #117, opened 2022-12-21
Fix broken tests


nkandpa2:
After #114, tests for `git_theta/params.py` and `git_theta/utils.py` are broken since they test functions that have been removed.

Issue #115, opened 2022-12-07
Hash parameters with an LSH (PLZ comment with thoughts)


blester125:
## The Issue

Currently we are hashing parameters based on passing their `.tobytes` value to a hash function (sha-1 atm). This works when parameters are **exactly** the same. This will start to cause issues when we begin to use parameter efficient updates. There are basically 2 places where this issue can arise:

1) Communication Efficient Update inference: When back-calculating things, (like a low-rank update based on the previous and current value via SVD) we can find update values that result in close parameters (i.e. they pass `np.allclose` even with stricter than default `rtol` and `atol` settings) that don't pass hash equality (i.e. `hash(current) != hash(previous + update)`). Depending on what parameter hash we store in the metadata file we will either have git reporting a diff right after a checkout or the hash in the metadata file won't actually match the hash of the parameters created during smudge. 
* This also opens the door for a systematic divergence between the checked in parameters and the working dir parameters. If the working dir parameters are never smudged then the each change will be from a different starting point and the communication efficient update will be calculated based on a different previous value than the training update was made with.

2) Communication Efficient Update application: The second place where you could get issues is in the application of updates on different machines i.e. when a smudge happens when checking out a repo on a new machine. Things that are non-deterministic can result in small amounts of noise to the point where it passes `np.allclose` but not hash equality. This will result in `git checkout && git status|diff` reporting changes in the checkpoint file.

## Solutions

`np.allclose` is non-transitive so we can't design a hash function where np.allclose equality implies hash equality and actually using `np.allclose` for equality checks will result in many unnecessary reads.

Instead we can use a Locality Sensitive Hash (LSH) to hash parameter values. LSH's are designed so that similar items hash together. Their main use-case is cheap similarity search based on the sketch (`LSH(real_value)`) but using that sketch as our hash bucket should result in hash equality when the parameter values are close to each other, even if they aren't an exact match. Most LSH's also have parameters that let you tune the how well they approximate the true similarity functions (at the cost of taking extra space) so we should be able to configure the LSH to be reasonable sure we can avoid collisions.

### Random Hyperplanes

[...]

PR #119, opened 2022-12-22
Fix broken tests and add test cases for new classes

nkandpa2:
Addresses #117


nkandpa2:
Some notes for this PR to consider:
- Tensorstore prior to release 0.1.14 has a different API for the `tensorstore.Tensorstore` object (no access to the `kvstore` associated with that object) that breaks git-theta. We cannot require a version >= 0.1.14 while also supporting python 3.6 since tensorstore >= 0.1.14 requires python >= 3.7. To fix this I just bumped our python requirement to >= 3.7. 
- The `test_tensor_metadata_machine_epsilon` in `tests/metadata_test.py` does not always succeed with the current method of doing `TensorMetadata` equality (i.e. round and then hash). This is related to #115 and will hopefully be fixed by that issue.

setup.py
```diff
@@ -61,11 +61,9 @@ def get_version(file_name: str, version_variable: str = "__version__") -> str:
     license="MIT",
     install_requires=[
         "GitPython",
-        "tensorstore",
+        "tensorstore >= 0.1.14",
         "file-or-name",
         "importlib_metadata",
```
blester125, line 66: With a bump to 3.7, we can remove the `importlib_metadata` backport too

nkandpa2: I think `importlib.metadata` was added to stdlib in 3.8 and was provisional until 3.10 based on https://docs.python.org/3/library/importlib.metadata.html and that's the way we seem to be treating it in `git_theta/updates/base.py`

blester125: You right I commented before I checked lol.

tests/theta_test.py
```diff
@@ -0,0 +1,105 @@
...
+
+import pytest
+import os
+import tempfile
+import git
+import random
+import string
+
+from git_theta import theta, git_utils
+from tests import testing_utils
```
blester125, line 11: I have never understood the best way to have importable code in tests. If this is the best way to do it just leave it, otherwise can you add a TODO (or an issue) to update to the best way (whatever pytest recommends lol)

tests/metadata_test.py
```diff
@@ -0,0 +1,152 @@
...
+
+import numpy as np
+import string
+import random
+import tempfile
+
+from git_theta import metadata
+
+
+def random_lfs_metadata():
```
blester125, line 11: Can some of these random data makers move into the testing util you added?

blester125 (changes requested):
Looks great! Added tests seem to cover a lot! Mostly cosmetic changes outside of the backport thing

git_theta/metadata.py
```diff
@@ -108,7 +93,7 @@ def from_metadata_dict(cls, d):
         return cls(metadata)
 
     @classmethod
-    @file_or_name
+    @file_or_name(file="r")
```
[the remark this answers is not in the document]
nkandpa2: Yeah, haven't looked into why this was needed too deeply, but I'm pretty sure since `cls` is the first argument `file_or_name` opens that instead of `file`.

```
In [8]: class A:
   ...:     @classmethod
   ...:     @file_or_name(f='w')
   ...:     def method(cls, f):
   ...:         print(cls, type(cls))
   ...:         print(f, type(f))
   ...:         f.write('abcd')
```

git_theta/metadata.py
```diff
@@ -21,24 +22,16 @@ class LfsMetadata(MetadataField):
...
+    name: ClassVar[str] = "lfs_metadata"
 
     @property
     def lfs_pointer(self):
-        return (
-            f"version {self.version}\noid sha256:{self.oid}\nsize {self.size}\n".encode(
-                "utf-8"
-            )
-        )
+        return f"version {self.version}\noid sha256:{self.oid}\nsize {self.size}\n"
```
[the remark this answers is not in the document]
nkandpa2: I changed this for consistency between the type returned by `lfs_pointer` and the type expected by `from_pointer`. Added the type annotations for this and the rest of this file

.github/workflows/unittest.yml
```diff
@@ -16,17 +16,7 @@ jobs:
...
-        # provides, so skip that configuration.
-        exclude:
-          - os: ubuntu-latest
-            python-version: "3.6"
-        # Python 3.6 is still in the ubuntu-20.04 container from github so
-        # test git-theta in that container instead.
-        include:
-          - os: "ubuntu-20.04"
-            python-version: "3.6"
+        python-version: ["3.7", "3.8", "3.9"]
```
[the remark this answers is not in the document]
nkandpa2: Test gets stuck trying to build tensor store for 3.11 so I'll make an issue.

git_theta/metadata.py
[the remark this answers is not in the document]
blester125: Thanks! Adding the arguments solves it here. I'll (eventually :grimacing:) get it fixed upstream lol.

tests/__init__.py
```diff
@@ -0,0 +1 @@
+from . import testing_utils
```
blester125, line 1: Does adding this `__init__.py` to the tests dir result in it getting picked up and installed as a package (called `tests`) based on `find_packages` used in the setup.py script? You can check by pip installing the package (no -e) and then seeing if a `tests` dir has shown up in your python site-packages dir (where this is depends on python install method.

tests/metadata_test.py
```diff
@@ -0,0 +1,103 @@
...
+from tests.testing_utils import (
+    random_lfs_metadata,
+    random_tensor_metadata,
+    random_theta_metadata,
+    random_param_metadata,
+    random_metadata,
+)
+
+
+def metadata_equal(m1, m2):
```
blester125, line 19: What is the output we get when these aren't equal? Pytest normally shows info about what doesn't match, is that available with this? If not can we add an issue to eventually have better integration with asserts for seeing what the mismatch is?

git_theta/utils.py
```diff
@@ -61,3 +62,33 @@ def unflatten(d: Dict[Tuple[str], Any]) -> Dict[str, Union[Dict[str, Any], Any]]
             curr = curr.setdefault(k, {})
         curr[ks[-1]] = v
     return nested
+
+
+def is_valid_oid(oid: str) -> bool:
```
blester125, line 67: Is there somewhere where we canonicalize hex hash values to be lower case? There are a few lower case assumptions sprinkled around the code base. 

Only working with lowercase is fine as long as we can enforce it if any tool we integrate with decides to switch to upper. Might make sense to add this as an issue instead of fixing it in this PR

tests/metadata_test.py
```diff
@@ -0,0 +1,103 @@
...
+        if m1_flat[k] != m2_flat[k]:
+            return False
+    return True
+
+
+def test_lfs_pointer():
+    """
+    Test LfsMetadata creates and reads LFS pointers correctly
+    """
+    version = "my_version"
```
blester125, line 34: Are `version` `oid` and `size` unused or am I missing something?

tests/params_test.py
```diff
@@ -1,67 +1,67 @@
...
+    """
+    Test TensorStoreSerializer serializes and deserializes correctly
+    """
+    serializer = params.TensorStoreSerializer()
+    for num_dims in range(1, 6):
+        shape = tuple(np.random.randint(1, 20, size=num_dims).tolist())
+        t = np.random.rand(*shape)
+        serialized_t = serializer.serialize(t)
+        deserialized_t = serializer.deserialize(serialized_t)
+        assert np.all(t == deserialized_t)
```
blester125, line 19: I think `np.testing.assert_array_equal` would probably be better for exact checks like this

tests/params_test.py
```diff
@@ -1,67 +1,67 @@
...
+
+
+def test_tensorstore_serializer_roundtrip_chunked():
```
blester125, line 22: I there a way to assert that this is actually partitioned by tensorstore? Like checking that there is more than just a `0.0` file?

tests/testing_utils.py
```diff
@@ -0,0 +1,74 @@
...
+    lfs_metadata = random_lfs_metadata()
+    theta_metadata = random_theta_metadata()
+    return metadata.ParamMetadata(
+        tensor_metadata=tensor_metadata,
+        lfs_metadata=lfs_metadata,
+        theta_metadata=theta_metadata,
+    )
+
+
+def random_metadata():
```
blester125, line 49: Lets do a quick refactor on this and the `make_nested_dict` utility.

1. Move `make_nested_dict` from `utils_test` to here.
2. Probably rename it `random_nested_dict`
3. Update the signature to take args like `possible_keys` and `possible_values` with defaults from the original `make_nested_dict` (convert to tuples or make the default values `None` and do a check in the function.
4. Update `random_metadata` to call `random_nested_dict` with the correct possible keys and values and then convert the returned value into a Metadata object.

git_theta/git_utils.py
```diff
@@ -12,11 +12,7 @@
 import shutil
 import filecmp
 import sys
-
-if sys.version_info < (3, 7):
```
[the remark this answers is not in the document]
blester125: I see that it is fixed in the code

blester125 (approved):
I listed some small changes but they can be punted to issues and added in other PR if you want

tests/__init__.py
nkandpa2: Good catch, that's exactly what happens in this case -- will remove the `__init__.py`.

tests/metadata_test.py
nkandpa2: Yup left it in by mistake.

tests/params_test.py
nkandpa2: My understanding is that when you aren't super opinionated about the chunking strategy Zarr tries to come up with what it thinks is the optimal chunking strategy for you. From testing on my system, I'd end up getting chunks for array dimensions with greater than 1000 elements IIRC.

PR #120, opened 2022-12-22
Update README.md

nkandpa2:
README was out of date so took a pass on it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,22 +19,22 @@ In order to track the model checkpoint with Git-Theta, run the command
 git theta track model.pt
 ```
 
-This command signals to Git (by adding to `.gitattributes`) that when running certain operations on `model.pt`, it should delegate to Git-Theta.
+This command configures Git to delegate to Git-Theta when performing certain operations on `model.pt`. This configuration can be seen in the `.gitattributes` file.
 
-With the model tracked by Git-Theta, you can treat the model checkpoint exactly like you would any other file in a Git repository. All of the regular Git commands like `add`, `commit`, `push`, `pull`, `checkout`, `status`, `diff`, etc. will work on your model the way they would on any other file.
+With the model tracked by Git-Theta, you can treat the model checkpoint exactly like you would any other file in a Git repository. All of the regular Git commands (`add`, `commit`, `push`, `pull`, `checkout`, `status`, `diff`, etc.) will work on your model the way they would on any other file.
 
 Additionally, when staging a change to a model, you can provide Git-Theta with additional information about ***what type*** of change is being staged (e.g., a sparse update, a low-rank update, etc.) by running `git theta add model.pt --update-type <update type>`. This allows Git-Theta to store the model update more efficiently, saving disk space and bandwidth when `push`-ing or `pull`-ing.
 
 # Why is this better than using Git or Git LFS?
-Git on its own can certainly be used for versioning non-text files like model checkpoints. However, the main limiting factors are that
+Git on its own can certainly be used for versioning non-text files such as model checkpoints. However, the main limiting factors are that
 1. Git remotes like Github and Bitbucket have a maximum file size (~50MB)
 2. Git is not designed to handle very large repositories
 
-There are a number of existing solutions for storing large files with Git that circumvent the maximum file and repository size, such as Git LFS. These work by pushing large files to an external LFS endpoint rather than the Git remote. The main issue with using Git LFS-like systems for versioning ML models is that they are unaware of the structure of ML models. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,101 +1,84 @@
...
 ```
 
-### A Single Deep Learning Framework
+This command signals to Git (by adding to `.gitattributes`) that when running certain operations on `model.pt`, it should delegate to Git-Theta.
 
-If you plan to track model checkpoints created by a single deep learning
-framework, for example only PyTorch or only Tensorflow, you can elect to only
-ensure the framework you use will be installed, avoiding the long install times
-and possible version requirements issues installing unused frameworks may bring.
+With the model tracked by Git-Theta, you can treat the model checkpoint exactly like you would any other file in a Git repository. All of the regular Git commands like `add`, `commit`, `push`, `pull`, `checkout`, `status`, `diff`, etc. will work on your model the way they would on any other file.
```
blester125, line 24: Possible re-write:

Regular Git command (`add`, `commit`, `push`, `pull`, `checkout`, `status`, `diff`, etc.) will...

README.md
```diff
@@ -1,101 +1,84 @@
...
 
-For example, install git-theta with only pytorch checkpoint support:
+Additionally, when staging a change to a model, you can provide Git-Theta with additional information about ***what type*** of change is being staged (e.g., a sparse update, a low-rank update, etc.) by running `git theta add model.pt --update-type <update type>`. This allows Git-Theta to store the model update more efficiently, saving disk space and bandwidth when `push`-ing or `pull`-ing.
```
blester125, line 26: Should we also mention it can be done with env vars and normal `git add`?

README.md
```diff
@@ -1,101 +1,84 @@
...
 
-``` bash
-cd git-theta
-pip install .[pytorch]
-```
+# Why is this better than using Git or Git LFS?
+Git on its own can certainly be used for versioning non-text files like model checkpoints. However, the main limiting factors are that
+1. Git remotes like Github and Bitbucket have a maximum file size (~50MB)
+2. Git is not designed to handle very large repositories
```
blester125, line 31: A few more reasons.

3. A small change to a single model parameter shows as a change to the while model checkpoint.
4. Diffs on non-text files do not help localize which model parameters changed.

Also maybe this could be earlier as part of the extra motivation you mentioned.

README.md
```diff
@@ -1,101 +1,84 @@
...
-If you already have your framework of choice installed (i.e. pip doesn't need
-to ensure it is installed), you can just install git-theta with `pip install .`
+There are a number of existing solutions for storing large files with Git that circumvent the maximum file and repository size, such as Git LFS. These work by pushing large files to an external LFS endpoint rather than the Git remote. The main issue with using Git LFS-like systems for versioning ML models is that they are unaware of the structure of ML models. 
 
-## Initializing git theta
-Initialize git theta by running:
-```bash
-git theta install
-```
+Imagine you have a checkpoint that you are updating by [training only a small percentage of the parameters](https://arxiv.org/abs/2111.09839), [training only a few of the layers](https://arxiv.org/abs/2106.10199), or by [adding new trainable modules](https://arxiv.org/abs/1902.00751). In these cases, most of the model remains the same and only a small fraction of the model gets modified. However, tools like Git LFS just see that the checkpoint file has changed, and will store the new version of the checkpoint file in its entirety.
```
blester125, line 35: We should probably update the links to papers from lab members that fit these categories. We can basically think of this as an advertisement ;)

README.md
```diff
@@ -1,101 +1,84 @@
...
-[filter "lfs"]
-        smudge = git-lfs smudge -- %f
-        required = true
-        clean = git-lfs clean -- %f
-[filter "theta"]
-        clean = git-theta-filter clean %f
-        smudge = git-theta-filter smudge %f
-        required = true
-```
+Git-Theta is aware of the structure of ML models and is designed to store only the parts of a model that have changed from its previous version as efficiently as possible.
```
blester125, line 37: `Git-Theta is aware of the structure of ML models`

I might be being too pedantic but I think we should reword to something like `ML models are comprised for multiple parameter groups and Git-Theta is designed to store...`

Because we don't actually know the structure (like if you are reusing embedding for not, etc) we just know it is made of multiple parameters that can change ~independently

blester125 (approved):
Looks good. suggested a few edits, feel free to ignore.

I think the comment about what "structure" means in this context is the only change we should def make

README.md
nkandpa2: I think details like this should be deferred to an actual tutorial/documentation page that we should make.

Issue #61, opened 2022-11-09
use tensorstore async for writing out parameter group files


craffel:

blester125:
Mixing async and normal code can be tricky. Here is the skeleton of a common way to do it (sorry if I over-explain).

1) Update the calls to tenorstore code to use the async versions of things (generally add an await and turn the function into an async def)
2) Make a write all function who's job is to start all of the async tasks
3) Execute this writer in a blocking way.

```python
async write_tracked_file(...):
  """This is the update to make the tensorstore write async."""
  
 async write_all_tracked_files(...):
  """This is what schedules the async tasks to be done concurrently."""
   await aysncio.gather(
       write_tracked_file(x) for x in ...
  )

# From your real code. This is executing it in a blocking way in the main synchronous code
asyncio.run(write_all_tacked_files(...))
```

If want to have a minimum python version of 3.7 we can just use this run, otherwise we probably need a custom runner like

```python
def async_run(to_run):
  loop = asyncio.get_event_loop()
  return loop.run_until_complete(write_all_tracked_files)
```

Let me know if anything is unclear

blester125:
The new update class modules changes how this will need to be implemented, we probably need to have `async` versions of the write and apply methods on those objects. We will also need to see if multiple async methods accessing different git commits causes contention, but effective parallelism/concurrency could be a major win

blester125:
This was done in #121

PR #121, opened 2022-12-25
Move to Async Read/Writes

blester125:
This PR updates the our clean and smudge filters to use async to do I/O bound operations (mostly tensorstore serialization and running git-lfs-(clean|smudge)). This is done with an async wrapper of subprocesses that mimics subprocess.run. It also includes tools for running an async function on all the k, v pairs in a dictionary so that they will happen concurrently.

The main async entry point is in the clean and smudge filters where the cleaning (or smuding) function is applied to each leaf in the parameter tree concurrently.

This also updates the Update API a little bit. This update makes async possible (i.e. some methods are converted from `def` to `async def`) and it also makes the API easier for writing new update plugins. Additionally the serializer is now passed to the update class and used internally instead of being used externally. Also things like communication with git-lfs is now done in the write method explicitly instead of being part of the LfsMeatadata creation.

There are still some change wrt to documentation, typing, and the exact Updater API but I wanted to get eyes on it sooner rather than later.

### DIFF
--- a/git_theta/git_utils.py
+++ b/git_theta/git_utils.py
@@ -7,18 +7,23 @@ import json
 import logging
 import io
 import re
-from typing import List, Union
+from typing import List, Union, Sequence
 import subprocess
 import shutil
 import filecmp
 import sys
 
+# TODO(bdlester): importlib.resources doesn't have the `.files` API until python
+# version `3.9` so use the backport even if using a python version that has
+# `importlib.resources`.
 if sys.version_info < (3, 9):
     import importlib_resources
 else:
     import importlib.resources as importlib_resources
 
 from file_or_name import file_or_name
+
+from git_theta import async_utils
 
 
 def get_git_repo():
@@ -240,26 +245,26 @@ def get_head(repo):
         return None
 
 
-def git_lfs_clean(file):
-    out = subprocess.run(
-        ["git", "lfs", "clean"], input=file, capture_output=True
-    ).stdout.decode("utf-8")
-    return out
+async def git_lfs_clean(file_contents: bytes) -> str:
+    out = await async_utils.subprocess_run(
+        ["git", "lfs", "clean"], input=file_contents, capture_output=True
+    )
+    return out.stdout.decode("utf-8")
 
 
-def git_lfs_smudge(pointer_file):
-    out = subprocess.run(
+async def git_lfs_smudge(pointer_file: str) -> bytes:
+    out = await async_utils.subprocess_run(
         ["git", "lfs", "smudge"],
         input=pointer_file.encode("utf-8"),
         capture_output=True,
-    ).stdout
-    return out
+    )
+    return out.stdout
 
 
-def git_lfs_push_oids(remote_name, oids):
+async def git_lfs_push_oids(remote_name: str, oids: Sequence[str]) -> int:
     if oids:
-        out = subprocess.run(
-            ["git", "lfs", "push", "--object-id", re.escape(remote_name)] + list(oids)
+        out = await async_utils.subprocess_run(
+            ["git", "lfs", "push", "--object-id", re.escape(remote_name)] + list(oids),
         )
         return out.returncode
     return 0
--- a/git_theta/params.py
+++ b/git_theta/params.py
@@ -1,5 +1,6 @@
 """Classes for serializing model updates."""
 
+from abc import ABCMeta, abstractmethod
 from collections import defaultdict
 import tensorstore as ts
 import io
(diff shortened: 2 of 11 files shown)


git_theta/file_io.py
```diff
@@ -0,0 +1,130 @@
+import tensorstore as ts
```
nkandpa2, line 1: This file was deleted in #114. Probably got added back into this branch inadvertently.

git_theta/git_utils.py
```diff
@@ -240,24 +245,24 @@ def get_head(repo):
...
+    return out.stdout
 
 
-def git_lfs_push_oids(remote_name, oids):
+async def git_lfs_push_oids(remote_name: str, oids: Sequence[str]) -> int:
     if oids:
-        out = subprocess.run(
-            ["git", "lfs", "push", "--object-id", re.escape(remote_name)] + list(oids)
+        out = await async_utils.subprocess_run(
+            ["git", "lfs", "push", "--remote", remote_name] + list(oids),
```
nkandpa2, line 265: Should the argument be `--object-id`?

git_theta/params.py
```diff
@@ -8,39 +9,37 @@
 
 
 class TensorSerializer:
```
nkandpa2, line 11: Should this also be an abstract class?

git_theta/updates/base.py
```diff
@@ -9,40 +10,148 @@
...
+    """Base class for parameter update plugins."""
+
+    def __init__(self, serializer: params.Serializer):
+        self.serializer = serializer
 
     @property
-    def name(self):
+    @abstractmethod
+    def name(self) -> str:
         raise NotImplementedError
```
nkandpa2, line 32: Is there any need to raise exception here since the method is abstract?

git_theta/updates/base.py
```diff
@@ -9,40 +10,148 @@
...
+        """Save and serialize (just) the update weights."""
+        serialized_update = await self.serializer.serialize({"parameter": update})
+        lfs_pointer = await git_utils.git_lfs_clean(serialized_update)
+        return metadata.LfsMetadata.from_pointer(lfs_pointer)
+
+    async def write(
+        self,
+        param: Parameter,
+        param_keys,
+        *,
```
nkandpa2, line 125: What's the purpose of the extra positional args here and in `apply`?

git_theta/updates/base.py
```diff
@@ -9,40 +10,148 @@
...
+        """Serialize and save a parameter with git-lfs."""
+
+    @abstractmethod
+    async def apply(
+        self, param_metadata: metadata.ParamMetadata, param_keys: Tuple[str], **kwargs
+    ) -> Parameter:
+        """Get the final parameter value, including fetching previous values."""
+
+
+class TrueUpdate(Update):
```
nkandpa2, line 54: What is your reasoning for the name `TrueUpdate`? To me something like `RecursiveUpdate` or `IncrementalUpdate` that alludes to the idea that the update depends on previous values seems more descriptive.

nkandpa2:
No issue for me with the direction of the `Update` API and async clean/smudge/push.

git_theta/file_io.py
blester125: I originally prototyped in this file and it got added back. Deleted

git_theta/git_utils.py
blester125: Fixed

git_theta/params.py
blester125: Good idea, updated

git_theta/updates/base.py
blester125: Yep, added docstring (as a body is required) and removed error. Did similar things elsewhere.

git_theta/updates/base.py
blester125: Updated to `IncrementalUpdate` good idea.

git_theta/updates/base.py
blester125: My thought here is that some updates (non-incremental, i.e. dense) are somewhat distinct from incremental updates (i.e. sparse).

What this lets us do is have the shared parameters of the API (`param` and `param_keys` in this case) as the only thing that is "needed" in the non-incremental updates. This `*` actually forces arguments that follow to be passed by keyword. Then when calling code passes the extra parameters for the incremental updates they have to pass the extra things via keyword and those parameters are absorbed by the `**kwags` on the base update.

This creates a visual distinction for developers between parameters that all updates will need and ones that only some will need.

Another option would be to have conditionals in the calling code that decides which params to pass or to have the base class take all the parameters, even if they are unused (but that seemed to make it hard to infer what an update would do).

nkandpa2: Nice, this makes sense!

blester125:
closes https://github.com/r-three/git-theta/issues/61

nkandpa2 (approved):

Issue #85, opened 2022-11-21
Add optional framework installs.


blester125:
As we have moved to plugins for checkpoint handling, we don't need all of the deep learning frameworks installed all the time. Therefore we don't need to install them all, especially given that they can be heavy.

Update the setup.py to include `extras_require` for various frameworks that install them with `git-theta`. Also include some target that installs all the frameworks, or at least some of the most popular ones.

PR #126, opened 2023-01-10
Adds a TF Checkpoint plugin.

blester125:
This PR adds a Tensorflow checkpoint plugin based on the checkpoint format output from `tensorflow.keras.Model.save_weights`. This format does not include the computational graph.

This PR includes tests for git-theta to read and write a tf checkpoint and ensure that the values are correct (one test leaves all values unchanged and another makes modifications).

It also leaves a skeleton for the tensorflow SavedModel checkpoint format but that is not implemented.

It also updates some of the installation so that not all frameworks are needed at once and it ensure that tests are only run of the required framework is installed.

Closes https://github.com/r-three/git-theta/issues/85
Closes https://github.com/r-three/git-theta/issues/79

### DIFF
--- a/git_theta/checkpoints/base.py
+++ b/git_theta/checkpoints/base.py
@@ -1,4 +1,4 @@
-"""Base class and utilities for different checkout format backends."""
+"""Base class and utilities for different checkpoint format backends."""
 
 from abc import ABCMeta, abstractmethod
 import os
--- a/git_theta/checkpoints/base.py
+++ b/git_theta/checkpoints/base.py
@@ -1,8 +1,7 @@
-"""Backends for different checkpoint formats."""
+"""Base class and utilities for different checkout format backends."""
 
+from abc import ABCMeta, abstractmethod
 import os
-import json
-import io
 import sys
 from typing import Optional
 import numpy as np
@@ -12,18 +11,16 @@ if sys.version_info < (3, 10):
 else:
     from importlib.metadata import entry_points
 
-from file_or_name import file_or_name
-
-from . import utils
+from git_theta import utils
 
 
-class Checkpoint(dict):
+class Checkpoint(dict, metaclass=ABCMeta):
     """Abstract base class for wrapping checkpoint formats."""
 
     @property
+    @abstractmethod
     def name(self):
         """The name of this checkpoint handler, can be used to lookup the plugin."""
-        raise NotImplementedError
 
     @classmethod
     def from_file(cls, checkpoint_path):
@@ -37,6 +34,7 @@ class Checkpoint(dict):
         return cls(cls.load(checkpoint_path))
 
     @classmethod
+    @abstractmethod
     def load(cls, checkpoint_path):
         """Load a checkpoint into a dict format.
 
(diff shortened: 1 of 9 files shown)


setup.py
```diff
@@ -72,13 +72,20 @@ def get_version(file_name: str, version_variable: str = "__version__") -> str:
...
+        # TODO: Is there a way for all to be the concat of the list of others
+        # instead of making a new list? In case some framework needs multiple
+        # libraries installed?
+        "all": ["torch", "tensorflow"],
     },
     entry_points={
         "git_theta.plugins.checkpoints": [
             "pytorch = git_theta.checkpoints:PickledDictCheckpoint",
             "pickled-dict = git_theta.checkpoints:PickledDictCheckpoint",
+            "tf = git_theta.checkpoints.tensorflow_checkpoint:TensorFlowCheckpoint",
```
nkandpa2, line 84: I guess many tf checkpoint aliases is good for now since users are manually setting the model type at the per-command level with an environment variable. I think we'd discussed moving away from this approach but that's another issue. 

Preliminary I was thinking that we should specify the model type when the model is first tracked with `git theta track <model_glob> --checkpoint-type <checkpoint_type>`. In addition to adding the filter, diff, and merge attributes for the `<model_glob>`, this could also add another attribute called `checkpoint_type`. Then every time we need to get the checkpoint handler, we just need to read the .gitattributes file and pick up the `checkpoint_type` attribute for the glob that matches the current file. The only times we call `get_checkpoint_handler` we always have the working-tree path to the checkpoint file so this should work. Another nice thing about this is that .gitattributes is part of the repo so the checkpoint type will be correctly set for anyone that clones an existing repo.

git_theta/checkpoints/base.py
```diff
@@ -1,8 +1,7 @@
-"""Backends for different checkpoint formats."""
+"""Base class and utilities for different checkout format backends."""
```
nkandpa2, line 1: Typo: "checkout" --> "checkpoint"

tests/tensorflow_checkpoint_test.py
```diff
@@ -0,0 +1,98 @@
...
+        # Load the Checkpoint
+        ckpt = tensorflow_checkpoint.TensorFlowCheckpoint.from_file(f.name)
+        with tempfile.NamedTemporaryFile() as f2:
+            # Use the git-theta save to create a new checkpoint
+            ckpt.save(f2.name)
+            # Load a model from they checkpoint we just saved
+            loaded_model = make_fake_model()
+            loaded_model.load_weights(f2.name)
+    for og, new in zip(fake_model.variables, loaded_model.variables):
+        np.testing.assert_allclose(og.numpy(), new.numpy())
```
nkandpa2, line 70: Should this be `assert_equal` here?

tests/checkpoints_test.py
```diff
@@ -69,4 +71,4 @@ def test_get_checkpoint_handler_pytorch(no_env_var):
     """Check that checkpoint_handler type is correct for when checkpoint_handler name resolves to pytorch"""
```
nkandpa2, line 71: This is the reason for the `pytest.importorskip` right? For a future issue we should make a `pickled_dict_checkpoint_test.py` with similar tests to what you wrote for tf and move this test into there.

nkandpa2 (approved):

git_theta/checkpoints/base.py
blester125: Fixed

tests/tensorflow_checkpoint_test.py
blester125: updated

tests/checkpoints_test.py
blester125: Yep, added a TODO

PR #128, opened 2023-01-11
Adding a sparse update type

muqeeth:


git_theta/updates/sparse.py
```diff
@@ -17,7 +19,18 @@ def name(self):
...
+            "indices": update.indices,
+            "indptr": update.indptr,
+            "shape": parameter.shape,
+        }
 
     async def apply_update(self, update: Parameter, previous: Parameter) -> Parameter:
-        return update + previous
+        param_update = scipy.sparse.csr_matrix(
+            (update["data"], update["indices"], update["indptr"]),
+            shape=(1, np.prod(update["shape"])),
```
nkandpa2, line 34: Should `shape` be `(np.prod(update["shape"]), )` instead of `(1, np.prod(update["shape"]))`? In `calculate_update` the update is flattened to be a 1-d array so seems like `shape` should also be 1-d. Probably doesn't make a difference in the result if the tests are passing.

nkandpa2 (approved):

git_theta/updates/sparse.py
blester125: The scipy sparse matrix always wants a 2d thing so it was erroring when the shape wasn't 2d.

We should add a comment here explaining that.

git_theta/updates/sparse.py
```diff
@@ -17,7 +19,18 @@ def name(self):
     async def calculate_update(
         self, parameter: Parameter, previous_parameter: Parameter
     ) -> Parameter:
-        return parameter - previous_parameter
+        diff = parameter - previous_parameter
+        update = scipy.sparse.csr_matrix(np.reshape(diff, (-1,)))
```
blester125, line 23: Does the dense -> csr transform do any sort of thesholding or is it only looking for actual 0's?

We should add a comment about what it does and think about if we want to add some sort of threshold for what we consider 0 (i.e. < 1e-12 could be converted to 0 to create a more sparse tensor).

tests/updates/sparse_update_test.py
```diff
@@ -0,0 +1,35 @@
...
+NUM_UPDATES = 100
+TRIALS = 50
+
+
+@pytest.fixture
+def updater():
+    return sparse.SparseUpdate(params.get_update_serializer())
+
+
+def test_sparse_update_application(updater):
```
blester125, line 21: probably rename this to something like round trip as it tests both the calculate update and the apply

blester125 (changes requested):
Just a few tweaks to add before we forget the details.

git_theta/updates/sparse.py
nkandpa2: Gotcha, sounds good

muqeeth: In `calculate_update` method, I changed the reshape to (1, flattened_shape) so that it is consistent with the `apply_update` method.

git_theta/updates/sparse.py
muqeeth: >We should add a comment here explaining that.

Added comment on why we need to give the shape for csr_matrix when reconstructing from data. 

>Does the dense -> csr transform do any sort of thesholding or is it only looking for actual 0's?

From what I see in documentation, it looks for actual 0's. There is no parameter to pass in a threshold value. If we plan to have a threshold, we can implement it on the update tensor before passing it to csr_matrix.  I will add a comment in the code on this.

tests/updates/sparse_update_test.py
muqeeth: Got it, renamed the function

blester125 (approved):

PR #133, opened 2023-01-13
sparse update with configurable threshold 

muqeeth:


git_theta/updates/sparse.py
```diff
@@ -12,6 +13,13 @@
 class SparseUpdate(IncrementalUpdate):
     """An update where only some parameters are touched."""
 
+    def __init__(
+        self, serializer: params.Serializer, threshold: Optional[float] = 1e-12
```
blester125, line 17: `thereshold` shouldn't be `Optional` it is always used.

blester125 (changes requested):
Can you add some quick tests? Things like:

* Given a consistent update value, increasing the threshold hold results in monotonically increasing sparseness
* Something that checks the sparseness of a known update? Like design one that will be 70% sparse after shareholding for example. 

You should be able to get sparseness ratios with `len(update_dict["values"]) / np.prod(param.shape)`

git_theta/updates/sparse.py
muqeeth: Changed it to always used

muqeeth:
Added above mentioned test cases

blester125 (approved):

PR #136, opened 2023-01-13
Fix bug because sparse tensor shape is a tuple.

blester125:
The TensorStore serializer only wants np.arrays (it calls methods on them). When it tried to call this on the shape tuple it failed.


muqeeth (approved):
Nice catch!

PR #137, opened 2023-01-16
LSH Comparison In Clean Filter

nkandpa2:
An initial cut at using Euclidean LSH to compute hashes for parameters and then using the hash values to efficiently check in the clean filter whether a parameter has changed *enough for the change to be tracked as a change*.

Implementation feels a bit messy so would like some thoughts on ways to improve it.


git_theta/updates/base.py
```diff
@@ -125,15 +125,19 @@ async def write(
...
+        prev_param_metadata: metadata.ParamMetadata,
         repo,
         path: str,
         **kwargs,
     ) -> metadata.LfsMetadata:
         """Serialize and save a parameter with git-lfs as a delta from the previous value."""
         logging.debug(f"Writing {self.name} update for {'/'.join(param_keys)}")
-        previous_value = await self.get_previous_value(
-            param_metadata, param_keys, repo=repo, path=path
+        prev_update_type = prev_param_metadata.theta_metadata.update_type
```
nkandpa2, line 135: @blester125 this fixes the bug where you need to do a second dense update before doing a sparse update.

git_theta/lsh/pool.py
```diff
@@ -0,0 +1,64 @@
...
+if sys.version_info < (3, 9):
+    import importlib_resources
+else:
+    import importlib.resources as importlib_resources
+
+
+# RandomnessPool is a singleton class so we don't read pool data off disk multiple times per git-theta command
+SINGLETON = None
+
+spec = [("pool", nb.float64[:]), ("index_hashes", nb.int64[:])]
```
blester125, line 20: Is there a reason to work in `int64` over `uint64`? Generally I think people work with uints when they are going to be doing bitwise things.

git_theta/lsh/pool.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+@nb.experimental.jitclass(spec)
+class RandomnessPool:
+    def __init__(self, pool, index_hashes):
+        self.pool = pool
+        self.index_hashes = index_hashes
+
+    def get_hyperplanes(self, feature_size):
```
blester125, line 29: I don't think you need two functions here, something like

```python
def get_hyperplanes(self, input_features, signature_idx):
  return self.pool[
    np.mod(np.bitwise_xor(input_features, self.index_hashes[signature_idx]), self.pool.size, dtype=np.int64)
  ] 
```

Then you can get all the hyperplanes at once with

```
pool.get_hyperplanes(
  np.expand_dims(np.arange(0, feature_size, dtype=np.int64)),
  np.arange(0, self.index_hashes.size, dtype=np.int64)
)
```

and a single element by just passing in the feature idx and the signature idx

git_theta/lsh/pool.py
```diff
@@ -0,0 +1,64 @@
...
+        pool_idx = int(np.mod(np.bitwise_xor(feature_idx, index_hash), self.pool.size))
+        return self.pool[pool_idx]
+
+
+def get_randomness_pool(signature_size):
+    global SINGLETON
+
+    if not SINGLETON:
+        package = importlib_resources.files("git_theta")
+        with importlib_resources.as_file(
```
blester125, line 49: Probably not needed before the paper but we should probably move this sort of thing into save/load methods on the hasher.

git_theta/utils.py
```diff
@@ -9,6 +9,8 @@
 class EnvVarConstants:
     CHECKPOINT_TYPE: str = "GIT_THETA_CHECKPOINT_TYPE"
     UPDATE_TYPE: str = "GIT_THETA_UPDATE_TYPE"
+    LSH_SIGNATURE_SIZE: str = "LSH_SIGNATURE_SIZE"
```
blester125, line 12: Unlikely anyone is, but we should prefix the actual env var with `GIT_THETA_*` to ensure we don't step on any toes.

git_theta/lsh/e2lsh.py
```diff
@@ -0,0 +1,89 @@
+"""Classes for computing Euclidean locality-sensitive hashes"""
```
blester125, line 1: I can't do a file level comment on the `.npy` files, but we should include a readme with info about what they are, i.e. why we have them, their shapes, their dtypes, the seed used to make them, etc.

blester125 (commented):
Left some comments asking about some aspects but I think it is close to merging.

git_theta/lsh/pool.py
```diff
@@ -0,0 +1,64 @@
...
+        for feature_idx in nb.prange(feature_size):
+            for signature_idx in nb.prange(self.index_hashes.size):
+                hyperplanes[feature_idx, signature_idx] = self.get_hyperplane_element(
+                    feature_idx, signature_idx
+                )
+        return hyperplanes
+
+    def get_hyperplane_element(self, feature_idx, signature_idx):
+        index_hash = self.index_hashes[signature_idx]
+        pool_idx = int(np.mod(np.bitwise_xor(feature_idx, index_hash), self.pool.size))
```
blester125, line 40: Needing this int was suspicious, I was playing around and found I could get different answers when my numbers were `np.uint64`s and I did a bitwise_xor followed by mod with an array as the second argument vs a scalar. When it is a vector I get the right answer (checked by hand) and the right type. When it is a scalar I get the wrong answer and a float return type.

I think numba defaults to unsigned ints for loop variables (I remember reading this but couldn't find it in the docs rn) and when that starts to mix with a scalar that somehow converts everything into floats and we get a floating mod instead of a int mod. The float mod seems to have more collisions in my spot checks too.

I think adding `dtype=np.uint64` (or a normal int64) fixes this in my quick checks.

bin/git-theta-filter
```diff
@@ -59,26 +68,43 @@ def clean(args):
...
         logging.debug(f"Cleaning {'/'.join(param_keys)}")
         param_metadata = prev_metadata.get(param_keys)
         new_tensor_metadata = metadata.TensorMetadata.from_tensor(new_param)
+
         # If the parameter tensor has not changed, just keep the metadata the same
-        # TODO: currently checks if rounded parameter hash has changed. Look into better parameter equality checks.
-        if param_metadata and param_metadata.tensor_metadata == new_tensor_metadata:
-            return param_keys, param_metadata
+        # TODO: Currently uses some magic numbers for LSH distance thresholds.
+        if param_metadata:
```
blester125, line 81: We probably want to do add shape and dtype checks here as a mismtach would let us skip any of the hash checking.

bin/git-theta-filter
```diff
@@ -59,26 +68,43 @@ def clean(args):
...
 
+    update_serializer = params.get_update_serializer()
     prev_metadata = metadata.Metadata.from_commit(repo, args.file, "HEAD").flatten()
 
     async def _clean(param_keys, new_param):
         logging.debug(f"Cleaning {'/'.join(param_keys)}")
         param_metadata = prev_metadata.get(param_keys)
         new_tensor_metadata = metadata.TensorMetadata.from_tensor(new_param)
+
         # If the parameter tensor has not changed, just keep the metadata the same
```
blester125, line 79: I think it is fine for now, but we might want to leave a todo about moving all this into the TensorMetadata class and then in the clean all we need to do is a `==` check.

bin/git-theta-filter
```diff
@@ -59,26 +68,43 @@ def clean(args):
...
-        if param_metadata and param_metadata.tensor_metadata == new_tensor_metadata:
-            return param_keys, param_metadata
+        # TODO: Currently uses some magic numbers for LSH distance thresholds.
+        if param_metadata:
+            hasher = lsh.get_lsh()
+            hash_distance = hasher.distance(
+                param_metadata.tensor_metadata.hash, new_tensor_metadata.hash
+            )
+            # If l2 distance < 1e-7, assume this change doesn't need to be tracked
+            if hash_distance < 1e-7:
```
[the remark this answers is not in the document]
nkandpa2: I've only tested this with the noise from by numpy (BLAS) non-determinism (i.e. different applications of the same sparse update result in slightly different arrays). `atol=1e-7` is sufficient for counting these as the same parameter.

bin/git-theta-filter
```diff
@@ -59,26 +68,43 @@ def clean(args):
     checkpoint_handler = checkpoints.get_checkpoint_handler()
     model_checkpoint = checkpoint_handler.from_file(sys.stdin.buffer)
 
+    update_serializer = params.get_update_serializer()
```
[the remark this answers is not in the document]
nkandpa2: Added

git_theta/lsh/pool.py
nkandpa2: I tried this out and it doesn't seem to work because numba doesn't support all types of advanced indexing, which is what's going on in `self.pool[...]` (https://numba.pydata.org/numba-doc/dev/reference/numpysupported.html)

bin/git-theta-filter
```diff
@@ -59,26 +68,43 @@ def clean(args):
...
     async def _clean(param_keys, new_param):
         logging.debug(f"Cleaning {'/'.join(param_keys)}")
         param_metadata = prev_metadata.get(param_keys)
         new_tensor_metadata = metadata.TensorMetadata.from_tensor(new_param)
+
         # If the parameter tensor has not changed, just keep the metadata the same
-        # TODO: currently checks if rounded parameter hash has changed. Look into better parameter equality checks.
-        if param_metadata and param_metadata.tensor_metadata == new_tensor_metadata:
-            return param_keys, param_metadata
+        # TODO: Currently uses some magic numbers for LSH distance thresholds.
```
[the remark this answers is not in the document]
nkandpa2: Moved the magic numbers into `utils.EnvVarConstants` so that they are configurable. Also to avoid having to make a getter for every environment variable that defaults to the correct value, I just built that into the `EnvVarConstants` class.

bin/git-theta-filter
```diff
@@ -59,26 +68,43 @@ def clean(args):
...
+                return param_keys, param_metadata
+            # If 1e-7 < l2 distance < 1e-5, load up old parameter and check with np.allclose
+            elif hash_distance < 1e-5:
+                update_handler = updates.get_update_handler(
+                    param_metadata.theta_metadata.update_type
+                )(update_serializer)
+                param = await update_handler.apply(
+                    param_metadata, param_keys, repo=repo, path=args.file
+                )
+                if np.allclose(param, new_param, rtol=0, atol=1e-7):
```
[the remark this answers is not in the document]
nkandpa2: I exposed both `rtol` and `atol` through `utils.EnvVarConstants` which I thought would be fine until we put some more thought into where this global repo-level configuration should go. 

Also, I kept it so that in the LSH signature distance check it still compares against `atol`. The reasoning is that the LSH signature distance approximates the L_2 distance between the two tensors and L_2 < L_inf. Thus if L_2 distance between two signatures is less than `atol`--> L_2 distance between the two tensors is likely to be less than `atol` --> L_inf distance between the tensors is less than `atol` --> `np.allclose(tensor1, tensor2, atol, rtol)` will pass.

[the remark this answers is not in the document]
nkandpa2: Good call, fixed

bin/git-theta-filter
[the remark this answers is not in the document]
nkandpa2: Ended up changing this to `atol=1e-8` and `rtol=1e-5` which matches the default values in `np.allclose`.

git_theta/lsh/base.py
```diff
@@ -0,0 +1,29 @@
+"""Base class for computing locality-sensitive hashes"""
+
+import abc
+from git_theta.lsh import get_randomness_pool
+from git_theta.lsh.types import Signature, Parameter
+
+
+class HashFamily(metaclass=abc.ABCMeta):
+    def __init__(self, signature_size: int):
```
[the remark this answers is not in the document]
nkandpa2: Talked to @craffel today and he pointed out that if you create a `np.random.BitGenerator` object by picking one of the classes that inherit from `BitGenerator` then you can get consistent randomness across systems. Instead of pre-computing and distributing `.npy` files, the random pool is initialized from a Mersenne Twister PRNG seeded with a fixed value. See `__init__` in `RandomnessPool`.

git_theta/lsh/e2lsh.py
```diff
@@ -0,0 +1,89 @@
...
+import numpy as np
+import numba as nb
+import os
+
+from git_theta.lsh import HashFamily, RandomnessPool
+from git_theta.lsh.types import Signature, Parameter
+from git_theta.utils import EnvVarConstants
+
+
+class E2LSH(HashFamily):
```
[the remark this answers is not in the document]
nkandpa2: Changed the name to `EuclideanLSH`. I think I see where this gets confusing because when using banding they refer to the banding step as the "LSH" right? I'm fine with either name. Left it as `EuclideanLSH` for now but happy to change it.

git_theta/lsh/e2lsh.py
```diff
@@ -0,0 +1,89 @@
...
+    Also see http://mlwiki.org/index.php/Euclidean_LSH for an introduction to E2LSH
+    """
+
+    def __init__(self, signature_size: int, bucket_width: float):
+        super().__init__(signature_size)
+        self.bucket_width = bucket_width
+
+    @property
+    def name(self) -> str:
+        return "E2LSH"
```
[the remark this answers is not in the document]
nkandpa2: Fixed

git_theta/lsh/e2lsh.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def hash(self, x: Parameter) -> Signature:
+        """Convert `x` to its signature."""
+        x = x.ravel()
+        hyperplanes = self.pool.get_hyperplanes(x.size)
+        return np.rint((x @ hyperplanes) / self.bucket_width).astype(np.int64).tolist()
```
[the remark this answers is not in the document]
nkandpa2: Good call, fixed

git_theta/lsh/e2lsh.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def distance(self, query: Signature, data: Signature) -> float:
+        """Compute the distance between two E2LSH signatures"""
+        query, data = np.array(query), np.array(data)
```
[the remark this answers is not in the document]
nkandpa2: Added the custom encoder and added a `TensorMetadata` `__post_init__` to turn the list back into a numpy array.

git_theta/lsh/e2lsh.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+@nb.jit(nopython=True, parallel=True)
+def nb_hash(
+    x: Parameter, signature_size: int, pool: RandomnessPool, bucket_width: float
+) -> Signature:
+    signature = np.zeros(signature_size)
+
+    for signature_idx in nb.prange(signature_size):
+        for feature_idx in range(x.size):
```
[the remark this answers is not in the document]
nkandpa2: Good call, supported by numba

git_theta/lsh/e2lsh.py
```diff
@@ -0,0 +1,89 @@
...
+            hyperplane_element = pool.get_hyperplane_element(feature_idx, signature_idx)
+            signature[signature_idx] += x[feature_idx] * hyperplane_element
```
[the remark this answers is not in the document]
nkandpa2: AFAIK `parallel=True` won't actually parallelize loops. Only operators like `dot`, `sum`, `prod`, etc. If you explicitly specify that a loop is parallelizable with `nb.prange` only then will it parallelize loops.

git_theta/lsh/e2lsh.py
```diff
@@ -0,0 +1,89 @@
...
+
+    return np.rint(signature / bucket_width).astype(np.int64)
```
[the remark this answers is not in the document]
nkandpa2: Fixed

git_theta/lsh/e2lsh.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def get_signature_size():
+    signature_size = os.environ.get(EnvVarConstants.LSH_SIGNATURE_SIZE)
+    return int(signature_size) if signature_size else 16
+
+
+def get_bucket_width():
+    bucket_width = os.environ.get(EnvVarConstants.LSH_BUCKET_WIDTH)
+    return float(bucket_width) if bucket_width else 1e-7
```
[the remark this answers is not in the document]
nkandpa2: Yeah, from testing the LSH distances it seems to lose resolution below the bucket width. Tied this parameter to `atol`.

git_theta/lsh/pool.py
```diff
@@ -0,0 +1,64 @@
...
+# version `3.9` so use the backport even if using a python version that has
+# `importlib.resources`.
+if sys.version_info < (3, 9):
+    import importlib_resources
+else:
+    import importlib.resources as importlib_resources
+
+
+# RandomnessPool is a singleton class so we don't read pool data off disk multiple times per git-theta command
+SINGLETON = None
```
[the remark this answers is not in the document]
nkandpa2: Since this class doesn't load data off disk anymore I just removed the singleton pattern

git_theta/lsh/pool.py
```diff
@@ -0,0 +1,64 @@
...
+
+spec = [("pool", nb.float64[:]), ("index_hashes", nb.int64[:])]
+
+
+@nb.experimental.jitclass(spec)
+class RandomnessPool:
+    def __init__(self, pool, index_hashes):
+        self.pool = pool
+        self.index_hashes = index_hashes
```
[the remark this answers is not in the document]
nkandpa2: Agreed, much better name. Fixed.

git_theta/lsh/pool.py
nkandpa2: From some testing (and you mentioned this below), `np.mod` of an unsigned type seems to return a float. To avoid this just used int64 everywhere.

git_theta/lsh/pool.py
nkandpa2: Removed singleton pattern since this class isn't loading data off disk anymore

git_theta/utils.py
nkandpa2: Fixed.

git_theta/lsh/e2lsh.py
nkandpa2: The files aren't needed anymore since we can consistently generate seeded random values

git_theta/lsh/pool.py
nkandpa2: Yeah I was seeing something similar when mod-ing with a scalar. It works if the first argument is `np.int64` but when it's `np.uint64` it returns a float. Just changed the arrays being fed into `bitwise_or` into `np.int64` arrays and it removed the issue.

bin/git-theta-filter
nkandpa2: Yeah I struggled with how to do this. Ideally the `==` check would compute the hash distance and if needed load the actual tensors and double check with `np.allclose`. This means that this should probably go in `__eq__` for `ParamMetadata` since this has access to the LSH hash as well as the LFS metadata needed to actually materialize the tensor. However, materializing the tensor (the way it's currently implemented with update handlers) requires some extra information like the file name, param keys, etc.

This is possible but will probably require some refactoring that we should push to another issue.

bin/git-theta-filter
nkandpa2: Added.

git_theta/lsh/pool.py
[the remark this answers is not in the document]
nkandpa2: Not sure how big an issue this is but I would prefer hashing the int since that wouldn't set a cap on the size of a parameter group

git_theta/lsh/e2lsh.py
```diff
@@ -0,0 +1,89 @@
...
+        for feature_idx in range(x.size):
+            hyperplane_element = pool.get_hyperplane_element(feature_idx, signature_idx)
+            signature[signature_idx] += x[feature_idx] * hyperplane_element
+
+    return np.rint(signature / bucket_width).astype(np.int64)
+
+
+def get_signature_size():
+    signature_size = os.environ.get(EnvVarConstants.LSH_SIGNATURE_SIZE)
+    return int(signature_size) if signature_size else 16
```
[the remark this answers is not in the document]
nkandpa2: From testing it seems like approximating the L_2 distance with just 16 random projections is pretty accurate. We also want to keep this as small as possible because we don't want to do too many projections of larger parameter groups (e.g., an embedding matrix of size 50000x1000 ~ 50M parameters). Running the non-numba LSH implementation on tensors of this size is extremely slow (at least on my laptop). With the numba implementation it takes about half a second to hash this tensor which isn't great but is usable.

tests/updates/base_test.py
```diff
@@ -6,10 +6,13 @@
 from git_theta import utils
 
 
+ENV_UPDATE_TYPE = "GIT_THETA_UPDATE_TYPE"
```
blester125, line 9: I there a reason for duplicating the string over using `utils.EnvVarConstants.UPDATE_TYPE.name?

setup.py
```diff
@@ -46,7 +46,14 @@ def get_version(file_name: str, version_variable: str = "__version__") -> str:
     author_email="craffel@gmail.com",
     url="https://github.com/r-three/checkpoint-vcs",
     packages=find_packages(),
-    package_data={"git_theta": ["hooks/post-commit", "hooks/pre-push"]},
+    package_data={
+        "git_theta": [
+            "hooks/post-commit",
+            "hooks/pre-push",
+            "lsh/data/pool.npy",
+            "lsh/data/index_hashes.npy",
```
blester125, line 54: Are these still needed with the new Generator code in the pool init?

git_theta/utils.py
```diff
@@ -4,11 +4,27 @@
...
 import re
+from dataclasses import dataclass
+
+
+@dataclass
+class EnvVar:
+    name: str
+    default: Any
+
+    def __get__(self, obj, objtype=None):
```
blester125, line 15: This is cool, I've never seen this before

blester125 (approved):
I think a few cleanups from recent changes got missed but other than that looks really good!

tests/updates/base_test.py
nkandpa2: So with that `__get__` method defined, any time you access a member of EnvVarConstants it automatically turns the accessed result into whatever `__get__` returns. So `utils.EnvVarConstants.UPDATE_TYPE` automatically turns into a string like `"dense"` so we lose the `.name` attribute.

blester125: Makes sense

Issue #24, opened 2022-09-28
Create binary for merging


craffel:
Probably it should just always designate a merge conflict? We also could eventually implement parameter averaging or allowing merges when complementary sets of parameters are updated.

craffel:
Depends on #52

craffel:
Brian's update: https://github.com/r-three/git-theta/issues/52#issuecomment-1379033545

PR #144, opened 2023-01-17
Git-Theta custom merge tool

blester125:
This PR adds a custom git-theta aware merge driver.

The driver loads all three versions of the (cleaned) checkpoint, branch A, branch B, and the merge-base (the ancestor commit for both). It then iterates through each parameter checking if it has changed.

When it finds a change it creates a prompt and collects user actions on what to do next. These actions are defined via a new plugin type. Additionaly these plugins register what strings they want to use as activation commands and what kind of merge situations they are valid for.

The prompt includes several qol things like type-ahead autocomplete, autosuggest, and input validation meaning that one cannot enter invalid commands.

closes https://github.com/r-three/git-theta/issues/24

### DIFF
--- a/git_theta/git_utils.py
+++ b/git_theta/git_utils.py
@@ -147,7 +147,7 @@ def write_gitattributes(
     gitattributes_file.write("\n")
 
 
-def add_filter_theta_to_gitattributes(gitattributes: List[str], path: str) -> str:
+def add_theta_to_gitattributes(gitattributes: List[str], path: str) -> str:
     """Add a filter=theta that covers file_name.
 
     Parameters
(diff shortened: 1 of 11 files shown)


git_theta/git_utils.py
```diff
@@ -174,11 +174,13 @@ def add_filter_theta_to_gitattributes(gitattributes: List[str], path: str) -> st
                 pattern_found = True
                 if not "filter=theta" in match.group("attributes"):
                     line = f"{line.rstrip()} filter=theta"
+                if not "merge=theta" in match.group("attributes"):
```
nkandpa2, line 177: Can we change this function name to `add_theta_to_gitattributes` since it's doing filters, merge drivers, and eventually diff drivers.

git_theta/merges/average.py
```diff
@@ -0,0 +1,80 @@
...
+        # Averaging with only one added parameter doesn't make sense.
+        DiffState.ADDED_A,
+        DiffState.ADDED_B,
+        # Averaging with a deleted parameter doesn't make sense
+        DiffState.DELETED_A,
+        DiffState.DELETED_B,
+        DiffState.DELETED_BOTH,
+    }
+
+    def average(self, a: Parameter, b: Parameter) -> Parameter:
```
nkandpa2, line 36: Can this be parameterized so that we can support interpolation

Can push this to another issue for later

nkandpa2 (approved):

git_theta/merges/average.py
blester125: This has been updated to work on variable number of params, but not interpolate yet. I did leave a TODO on adding interpolation tho.

git_theta/git_utils.py
blester125: Done

Issue #103, opened 2022-11-28
Update plugin names from object to class properties


blester125:
Current things like checkpoint or update plugins have a `.name` property that is object based, meaning it is only available once the object has been instantiated.

These names aren't instance specific, they are class specific (they are used to lookup the class in the plugins entry point) so they should be updated to be available before object instantiation.

Possible fixes:
* Find/Create something like `@property` for class based attributes
* Add `@classmethod` to the properties (unsure if the two decorators play nicely with each other)
* Update to just be class attributes (i.e. an assignment in the class body, not a method). The problem with this is we can't enforce subclasses to set it the way we can with an abstract method. One way to deal with that would be to `__subclass_init__` and metaclasses with reflection to automatically create a name if they provide it.

PR #180, opened 2023-01-30
Add abstract Class Attributes

blester125:
The PR changes the `name` attribute on plug-ins from being properties to being class attributes. This makes it possible to access information about the plug-in (name etc) without having to instantiate it.

It also adds a new class decorator that allows us to enforce the defining of an attribute in subclasses (i.e. user plugins), similar to the way we can for object-level properties.

### DIFF
--- a/git_theta/checkpoints/base.py
+++ b/git_theta/checkpoints/base.py
@@ -14,13 +14,11 @@ else:
 from git_theta import utils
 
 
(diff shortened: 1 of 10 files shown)


blester125:
closes https://github.com/r-three/git-theta/issues/103

nkandpa2 (approved):

Issue #138, opened 2023-01-17
Move from `scripts` to `entry_points['console_scripts']`


blester125:
We should move our binaries to be console scripts instead of using the `scripts` parameter.

[The Docs](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#console-scripts) say that this is more reliable for cross-platform.

Additionally using should result in the updated binary being used when code is changed during a `-e` install.

blester125:
They may need to be moved into a submodule like `git_theta/scripts/...` or something

PR #181, opened 2023-01-30
Move scripts to entry points.

blester125:
This PR updates the binaries from git-theta to live in `git_theta/scripts/...` instead of `bin/...` and uses the `console_scripts` entrypoint instead of the `scripts` argument.

The former is the recommended method and is supposed to smooth over some possible windows issues. Additionally this makes it easier to make changes to the runnable files in an editable install (you no longer need to re-run pip install after a change).

There is no user-facing change as the commands `git-theta`, `git-theta-filter`, etc. still work (they are the entry point names).

closes https://github.com/r-three/git-theta/issues/138

### DIFF
--- a/git_theta/scripts/git_theta.py
+++ b/git_theta/scripts/git_theta.py
@@ -158,7 +158,7 @@ def add(args, unparsed_args):
         repo.git.add(*unparsed_args)
 
 
-if __name__ == "__main__":
+def main():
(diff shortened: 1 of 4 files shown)


nkandpa2 (approved):

nkandpa2:
We can get rid of the shebang lines at the top of each script, right? Other than that LGTM

Issue #157, opened 2023-01-25
IA3 Update type


blester125:
Basically needs to infer the scaling vector in a subclass of `IncrementalUpdate.calculate_update`.

Might be a bit trick in deciding if it should just look at the last dim or the last few? (FF vs Attn?) Or might be easy to just use `weight.shape[1:]`? Generally all non leading dims are output dims? is that a safe assumption?

PR #185, opened 2023-02-08
IA3 update

muqeeth:
pull request for ia3 update as in #157


git_theta/updates/ia3.py
```diff
@@ -0,0 +1,38 @@
...
+    @property
+    def name(self):
+        return "ia3"
+
+    async def calculate_update(
+        self, parameter: Parameter, previous_parameter: Parameter
+    ) -> Parameter:
+        # since IA3 scales activations, weight.T[1:].shape can be considered as output dimensions and the update scalar vector/ matrix will be of shape = weight[1:].shape
+        # we transpose parameter weights so that output dimensions are in the end
+        parameter = parameter.T
```
blester125, line 27: This assumption about where the output dims are is pytorch specific, a tf implementation has them as the trailing dim by default.

Probably needs an argument about which dims are the broadcast dims? Or it might be easier to have the argument be which dims have ia3 applied.

Also in T5X there are cases where there are multiple ia3 dims (as attention weights are stored as 3d tensors) so we probably need them to be tuples and adjustable per parameter?

blester125 (changes requested):
I think there needs to be some updates to support other frameworks

git_theta/updates/ia3.py
```diff
@@ -0,0 +1,38 @@
...
+        previous_parameter = previous_parameter.T
+        # take into account divide by zero
+        multiplier = np.divide(
+            parameter, previous_parameter, where=previous_parameter != 0
+        )
+        update = np.mean(multiplier, axis=0, keepdims=True)
+        # Can we check if update == mulitplier[i] for any i?
```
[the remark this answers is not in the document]
muqeeth: Hmm, well I thought averaging over multiple vectors would smooth the noise. However, I am not sure if that happens in practice.

blester125:
The PR for side load information is [here](https://github.com/r-three/git-theta/pull/201) to help you update the Update class. It might be as simple as adding the `required_keys` class attribute.

It would also be nice to add an e2e test similar to the sparse/low-rank ones. (would need to run `./make-test.sh` and then update the `test.sh` file. Also probably add to the `model.py` script to add an option for generating test ia3 parameters).

LMK if you have any issues.

muqeeth:
Sounds good. I was about to ask you about IA3. It looks like the side loaded update for IA3 is missing right now. I will look into it based on the other update formats.

tests/end2end/model.py
```diff
@@ -63,6 +75,16 @@ def low_rank_update(t, rank):
     return {"A": R, "B": C}
 
 
+def make_ia3_update(value):
+    ia3 = np.random.randn(*value.shape)
+    if ia3.ndim > 3:
```
blester125, line 80: Probably simpler to have this section be something like

```python
axes = (0, -1) if ia3.ndim > 3 else (-1,)
ia3 = np.mean(ia3, axis=axes, keepdims=True)
```

muqeeth: Thanks for your suggestion, I made the modification. 

Can you help me with the following?

The below code in test.sh breaks for IA3 
```
echo "Verify that 'git status' doesn't have a diff."
git diff-index --quiet HEAD --
if [[ "$?" != 0 ]]; then
    exit 1
fi
```
The problem is that `git status` shows the file model.pt as modified after checking out to main in the test.sh file. I am not sure how to fix this issue.

Also, I noticed that if I remove the --quiet argument, we can't catch the error and the test.sh file says that the IA3 test is passed. Do you have any suggestions on how to handle this?

tests/end2end/model.py
```diff
@@ -63,6 +75,13 @@ def low_rank_update(t, rank):
     return {"A": R, "B": C}
 
 
+def make_ia3_update(value):
+    ia3 = np.random.randn(*value.shape)
```
blester125, line 79: You need to add `.astype(np.float32)` here, right now these are float64 so when the final diff is done the git repo has things saved as float32 but the file on disk has them saved as float64

tests/end2end/model.py
```diff
@@ -125,6 +144,26 @@ def main(args):
...
+    elif args.action == "ia3":
+        previous = torch.load(args.previous)
+        ia3_update = {name: make_ia3_update(value) for name, value in previous.items()}
+        update_handler = git_theta.updates.get_update_handler("ia3")
+        # Just the ia3 data
+        update_data = {}
+        # The updated parameter values
+        new_model = {}
+        for name, update in ia3_update.items():
+            if isinstance(update, dict):
```
blester125, line 156: You don't need this if check, all your updates are dicts.

tests/end2end/model.py
blester125: See https://github.com/r-three/git-theta/pull/185/files#r1169194198 on how to fix it.

Don't remove the `--quiet` argument, it's there so we can look at the exit code to see if it failed lol

muqeeth: hahaha, I removed to see the error message. But it turns out removing `--quiet` escapes the if condition :)

tests/end2end/ia3/clean.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
blester125, line 1: We should be able to remove this file as it is the same as the base clean file in the parent directory.

blester125 (approved):

PR #188, opened 2023-02-17
Feat/e2e tests

blester125:


.github/workflows/end2endtest.yml
```diff
@@ -0,0 +1,34 @@
...
+      - main
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      max-parallel: 4
+      matrix:
+        os: [ubuntu-latest, macos-latest]
```
craffel, line 18: IIRC we found that git-theta should work fine on windows as long as it wasn't using FAT32 and trying to track giant files, right? Any reason not to include windows-latest here?

tests/end2end/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+The command `make-test.sh "${test name}"` can be used to create the skeleton of a new test.
+
+### `test.sh`, the (Vegan-)?Meat and Potatoes
+
+The first steps of a `test.sh` file include sourcing the `../utils.sh` so it can use some of our shared functions. Then it should run `set -e` to ensure that errors in part of the test cause the whole test to fail. It should also call `test_init`, which will create a git repo (with a `main` branch) for it and ensure that `set -e` was used.
+
+The next step is often to create and modify some model. The provided `../model.py` file can be used as a helper to create new models and updates with special forms. This script creates two copies of the model, one that lives in the path that is version controlled and one that include information about how it was created. This path is returned on stdout, can be captured in `test.sh`. This checkpoint is not version controlled, uses the native checkpoint format, and should be removed by the clean script.
```
craffel, line 21: - that include*s* information
- *and* can be captured
- What do you mean by the "native" checkpoint format?

tests/end2end/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+The provided `verify.py` can be used to help check that version controlled models match they original values.
```
craffel, line 23: *their* original values

tests/end2end/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+`../utils.sh` provides a `commit` function that will create a git commit and return the hash for that commit, for example `SHA=$(commit "commit msg")` will make a commit and save the hash to `${SHA}`. It can be called with just `commit "commit msg"` if you do not want to track the hash. This should help for tests that travel through git history.
+
+> **Warning**
+> The `set -e` option causes a whole bash script to exit if one of the subcommands in it error/exits with a non `0` return code. Without this setting, tests that have failing steps will often appear as passing so we require this to be active. If you test has subcommands that are expected to fail and your test accounts for that correctly, you can suppress this `-e` behavior by adding `${subcommand} || true` to your failing commands (or call `set +e` after the `test_init` call). Both of **these settings can cause tests to fail silently**, so only use them if you are sure you understand what they are doing.
```
craffel, line 28: - in it *has* an error
- if *your* test

tests/end2end/commit/test.sh
```diff
@@ -0,0 +1,35 @@
...
+python ../verify.py --old-model "${MODEL_NAME}" --new-model "${INIT_MODEL}"
+
+echo "Making sure ${MODEL_NAME} is committed."
+FILES=$(git ls-files)
+if [[ ! ${FILES} =~ ${MODEL_NAME} ]]; then
+    red_echo "${MODEL_NAME} not found in 'git ls-files'."
+    exit 1
+fi
+
+green_echo "git commit test passed!"
```
craffel, line 35: Would it make sense to combine this script with the checkout script? It seems all actions done here (apart from the git ls-files check) are done in checkout implicitly.

tests/end2end/make-test.sh
```diff
@@ -0,0 +1,33 @@
...
+
+test_init
+EOF
+
+cat <<EOF > clean.sh
+#!/usr/bin/env bash
+rm -rf .git > /dev/null 2>&1
+rm -rf .gitattributes > /dev/null 2>&1
+rm *.pt > /dev/null 2>&1
+EOF
```
craffel, line 31: If clean.sh is the same for all tests, should we just put it in tests/end2end and have it called one level up?

tests/end2end/make-test.sh
```diff
@@ -0,0 +1,33 @@
...
+
+set -e
+
+test_init
+EOF
+
+cat <<EOF > clean.sh
+#!/usr/bin/env bash
+rm -rf .git > /dev/null 2>&1
+rm -rf .gitattributes > /dev/null 2>&1
```
craffel, line 29: Should we also remove .gitignore?

tests/end2end/model.py
```diff
@@ -0,0 +1,100 @@
...
+
+def main(args):
+    # Set seeds and enable determinism
+    random.seed(args.seed)
+    torch.manual_seed(args.seed)
+    np.random.seed(args.seed)
+    torch.use_deterministic_algorithms(True)
+
+    file_name, ext = os.path.splitext(args.model_name)
+    persistant_name = f"{file_name}-{args.action}-{args.seed}{ext}"
```
craffel, line 70: nit - persistent

tests/end2end/verify.py
```diff
@@ -0,0 +1,47 @@
...
+parser = argparse.ArgumentParser(description="Compare checkpoints for testing.")
+parser.add_argument("--new-model", help="The path to the new model.")
+parser.add_argument(
+    "--old-model", help="The path to the model we are comparing aganist."
+)
+parser.add_argument("--compare")
+
+
+def get_compare_function(compare):
+    """Eventually add configurable comaprison functions?"""
```
craffel, line 16: comparison

tests/end2end/verify.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def _cmp(a, b):
+        return np.allclose(a.numpy(), b.numpy())
```
craffel, line 19: Is allclose appropriate for all tests here? I'd think if we were checking e.g. that the checked-out version matches the original version, we can enforce equality, right? Since the data should be byte-for-byte equivalent using tensorstore?

craffel (changes requested):

.github/workflows/end2endtest.yml
blester125: I tried to add windows to the unittest a while ago and installing pytorch took so long that it kept timing out.

Also in this case the test scripts are all written in bash so I don't know if that would work on windows (we would need to use the linux subsystem somehow?)

craffel: I might be wrong but I think if we specify `shell: bash` (and resolve the pytorch install timeout) we can use windows-latest. bash is supported by all platforms; windows-latest uses powershell by default but that can be overriden. https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell We can deal with this in a follow-up though.

tests/end2end/README.md
blester125: Fixed

native means deep-learning framework specific, added that

tests/end2end/README.md
blester125: fixed

tests/end2end/README.md
blester125: fixed

tests/end2end/make-test.sh
blester125: `clean.sh` can be different for different test (i.e. in smuge is also deletes some json files) so currently we need to have per-test scripts. We could make the runner check for the existence of a test-specific but the clean commands are so small it doesn't seem like a big deal right now.

tests/end2end/make-test.sh
blester125: fixed

tests/end2end/verify.py
blester125: fixed

tests/end2end/commit/test.sh
blester125: Having both tests should help find issues if things ever break, i.e. if the checkout test breaks and this one doesn't we know the issues is in the smudge step, if they both break we know it is in clean, etc.

Issue #132, opened 2023-01-13
Include Commit Information in the Merge Tool UI


blester125:
Will help users remember who is A vs B etc.

blester125:
It seems like this is difficult/unreliable to do in the merge *driver* (decides how to actually merge the file) but is possible to do in a merge *strategy* (decides what needs to be merged).

[The CLI args for the merge strategy](https://stackoverflow.com/questions/47454227/how-to-configure-a-merge-strategy/47462046#47462046) seem to have all the commit info we want/need (which can be used to get other things like the first line of the commit message), but the job of a merge strategy is complicated.

My plan is to make a strategy that will collect all the cli arguments, store them in env vars, and then call the real merge strategy. Then the merge driver can read the values out of the ENV var.  Using ENV vars like this is already how the strategy normally gets this information, but [we don't have access to all the envs we need](https://stackoverflow.com/questions/24302644/how-to-retrieve-branch-names-in-a-custom-git-merge-driver) by the time we hit the driver. Reusing the env var mechanism makes sense tho.

Big questions are:
* How can we proxy to the default strategy
* Can we set the strategy from `.gitattributes`? All examples I have seen online include cli args to `git merge`

PR #190, opened 2023-02-17
Include some commit information in the merge ui

blester125:
This PR adds some information in the merge UI, it uses the `GITHEAD_${SHA}` env variable to tell what the other branch is and the git repo object for information about the current commit.

This summary is shown at the beginning of merging and can be shown again with the new "context" merge plugin.

This also adds the ability to have "no-op" merge plugins that do not advance the current parameter. Instead they do something like print the context and then re-display the menu of actions.

Below we can see the context menu shown at the start and as a result of the context plugin action
image
(You can see a bit of another terminal on the lower left, so the weird extra letters won't actually show up. Sorry)

closes https://github.com/r-three/git-theta/issues/132

### DIFF
--- a/bin/git-theta-merge
+++ b/bin/git-theta-merge
@@ -14,7 +14,7 @@ from prompt_toolkit.auto_suggest import AutoSuggestFromHistory
 from prompt_toolkit.completion import WordCompleter
 from git_theta import metadata
 from git_theta import merges
-from git_theta.utils import DiffState, Trie, TEXT_STYLE
+from git_theta.utils import DiffState, Trie, TEXT_STYLE, NoResult
 
 
 logging.basicConfig(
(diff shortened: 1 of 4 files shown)


nkandpa2 (approved):

nkandpa2:
Not related to this PR but is "average-ao" supposed to be "average-oo" with "oo" standing for "ours" and "original", or does the "a" stand for something else?

nkandpa2:
I think I get it but the naming isn't consistent. For average-ours-theirs, ours is "a" and theirs is "b", for average-theirs-original, theirs is "t", and for take_us, ours is "u". We should just use a consistent shorthand for ours, theirs, and original. I'd be fine with ours == "a", theirs == "b", and original == "o" or we could do ours == "o", theirs == "t", ancestor == "a".

PR #197, opened 2023-04-04
Add docs and doc building action

### LINKED ISSUE #77: Add documentation compilation and integration testing

craffel:
Fixes #77


craffel:
I fixed the doc building action. It now runs successfully, so we can now hopefully see if anyone writes a docstring with a syntax error automatically.

.github/workflows/build_docs.yml
```diff
@@ -0,0 +1,30 @@
...
+    - uses: actions/checkout@v2
+    - name: Set Up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: "3.8"
+    - name: Install Dependencies and Package
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install .[all,docs]
+        python -m pip install sphinx
```
blester125, line 26: redundant with the new `docs` target?

docs/source/conf.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+# -- Project information -----------------------------------------------------
+
+project = "git-theta"
+copyright = "2023, r-three"
+author = "r-three"
+
+# The full version, including alpha/beta/rc tags
+release = "0"
```
blester125, line 26: When we actually use these docs I think we should be able to import git_theta here and have the release number be the actually version number.

blester125 (approved):

.github/workflows/build_docs.yml
craffel: Done.

docs/source/conf.py
craffel: Done.

PR #198, opened 2023-04-04
Add Windows target for tests

### LINKED ISSUE #152: Ensure things work on windows.
When running experiments I got this warning from git-lfs 

```
Encountered 1 file that may not have been copied correctly on Windows:
	pytorch_model.bin

See: `git lfs help smudge` for more details.
```

I assume that because we are using git-lfs-smudge under the hood we are also seeing this warning but it is not being surfaced. Check if it is and if it is we should propagate it to the user.

craffel:
Might fix #152


craffel:
FYI @blester125 - https://github.com/r-three/git-theta/pull/198/commits/13436fb0b758ac5f59bd41f2467b578464ffe9ce changes the hardcoded /tmp path (which is not valid on Windows) used for logging to use tempfile.gettempdir instead. On Linux it will still use /tmp but on Mac it will use a process-specific path (like '/var/folders/pg/g4dpl9zn6tnc6ybgq7r158kr0000gn/T'). This will make the log file harder to find, but the process-specific path is always stored in the $TMPDIR environment variable. I assume that since you added this code and are on Linux you won't really care, but just FYI.

blester125 (approved):

blester125:
I added `print(psutil.Process().open_files())` to the `git_repo_with_commits` before the yield and then after the yield (both inside and outside side of the TempDir contexet manager)  fixture and then forced a test that uses it to fail and I'm seeing this in the output

```
----------------------------------------------------- Captured stdout setup ------------------------------------------------------
[]
---------------------------------------------------- Captured stdout teardown ----------------------------------------------------
[]
[]
```

So it seems like we aren't leaving any files open, i.e. this is some weird Windows things and not something we are doing wrong.

I also looped over all the processes that are open and none of them have files open that we use.

I also opened a file and then had the code above and it showed that open file so this code should detect any open files.

craffel:
Reading over https://stackoverflow.com/questions/65122957/resolving-new-pip-backtracking-runtime-issue, the recommended seems to basically be to pin more of our dependencies so that if it does end up backtracking it doesn't need to backtrack so much. If we don't want to be arbitrarily strict about dependencies that we don't need to be strict about, we could consider using a special requirements.txt file that is used during testing that pins all the dependencies. Using pipreqs, I get
```
file_or_name==1.1.6
flax==0.6.8
GitPython==3.1.31
importlib_metadata==4.11.1
importlib_resources==5.1.4
numba==0.56.4
numpy==1.23.5
prompt_toolkit==3.0.18
pytest==6.2.3
scipy==1.10.1
setuptools==49.2.1
six==1.15.0
tensorflow==2.12.0
tensorstore==0.1.27
torch==1.8.1
typing_extensions==3.7.4.3
```
Not sure what the best approach is here.

craffel:
It's failing because numpy 1.23.5 only supports Python 3.8+. Python 3.7 has end-of-life in on 2023-06-27, so I'm fine with dropping it (just pushed such a commit to make sure it works).

craffel:
Ok, tests are passing now. The problem is that this PR now does a few things:
1. Add testing for windows
2. Fix things so that they work on Windows
3. Remove support for 3.7
4. Add stricter requirements for tests

Should we break it up or just squash it all? @blester125

craffel:
It's not supported by numpy 1.23.5. We can lower the numpy version requirement. I think since it's almost EOL it's fine to drop.

PR #206, opened 2023-04-11
Update sparse_update_test.py

blester125:
Sometimes windows outputs 0.30001 instead of 0.30000 but it is def close enough for this test. Should fix that flakiness.


muqeeth:
Thanks for the fix. It seems windows is making our job harder.  Our test are run on Windows using Github actions after creating pull-request right? Why do think the tests did not fail in the beginning when I created sparse updates class?

muqeeth (approved):
Thanks for the fix. It seems windows is making our job harder.  Our test are run on Windows using Github actions after creating pull-request right? Why do think the tests did not fail in the beginning when I created sparse updates class?

muqeeth:
Thanks, understood.

Issue #210, opened 2023-04-21
User-facing memory ↔ git-theta save/load methods


blester125:
# Git Theta for Tracking Models Over a Training Run.

## Naive Implementation and Extra IO issue

We talked about the use case of using git-theta to track an experiment over the course of training. The simplest way to do this is like:

```python
m = BriansSuperCoolModel()
...
for batch in dataset:
  train_step(m, ..)
  ...
  if perf > best_perf:
    torch.save(m.state_dict(), "path/to/ckpt")
    sha = git_theta.commit("path/to/ckpt")
    best = sha
  ...
  
git_theta.checkout(best)
m = torch.load("path/to/ckpt")
run_eval(m)
```

We discussed ways to make things like committing non-blocking and hide that from the user, but the main issue we though of was that there are now extra reads and writes. Saving in now a torch native write in the training loop and a torch native read 
and git-theta write in the commit. Similarly loading is now a git-theta read and a torch native write in the checkout and a torch native read in the training loop.

## Git Theta Direct Storage Solution

A solution we had is to have direct memory $\iff$ git-theta transfer to avoid these double writes. We would create user facing functions like `git_theta.save|load` (or better names :woman_shrugging:?) and the loop would look something like this.

```python
m = MuqeethsEvenCoolerModel()
...
for batch in dataset:
  train_step(m, ...)
  ...
  if perf > best_perf:
    best = git_theta.save(m.state_dict(), "path/to/ckpt")
  ...

m = git_theta.load(best)
run_eval(m)
```

The advantages of this is that we would skip the extra reads and writes by not materializing the checkpoints on the file system.

### Efficient partial updates

We think this approach would also enable efficient storage of partial model updates without having to write unchanged parameters. For example, if only the decoder is updated in an EncDec model, when we call `git_theta_save` the parameter equality checks would make it so that only the decoder weights are added as git-lfs objects.

## Open Questions

* Should `git_theta.save` commit the model our just add it?
    * What happens to the working tree when you programatically make a commit? Does it get updated or do you end up in a detached HEAD state?
        * **The repo will be at the new commit which is now HEAD, but the file that was committed will not be in the working tree so a `git status` will show it as deleted.**
            * This means we may want to add a `git_theta.load(..., checkout=True)` flag that will actually smudge the model to disk? Then if they use this at the end of training they will have a model on disk and the repo will be at the commit for that model?
* How do we make this non-blocking and hide it from the user?
[...]

PR #211, opened 2023-04-21
A first crack at process memory <=> git-theta storage

blester125:
Demo:

```python
import os
import torch
import git_theta
from git_theta.api import save, load
import numpy as np

def make_model():
  return torch.nn.Sequential(
      torch.nn.Linear(10, 5),
      torch.nn.ReLU(),
      torch.nn.Linear(5, 2))

model = make_model()
model2 = make_model()

path = "model.pt"
sha = save(model.state_dict(), path, "model from py")

assert not os.path.exists(path)
assert not np.allclose(model[0].weight.detach().numpy(), model2[0].weight.detach().numpy())

m = load(sha, path)
model2.load_state_dict(m)

assert np.allclose(model[0].weight.detach().numpy(), model2[0].weight.detach().numpy())
assert not os.path.exists(path)
```

both asserts pass meaning we were able to save and load the model without it actually being written into the working tree!

First approximation of https://github.com/r-three/git-theta/issues/210


git_theta/api.py
```diff
@@ -0,0 +1,71 @@
...
+import git_theta
+from git_theta import checkpoints, filters, git_utils, metadata, utils
+
+
+# TODO: Should we have a `materialize` flag that writes the model to disk so you
+# don't have git status showing a diff?
+def save(
+    state_dict,
+    path: str,
+    commit_msg: Optional[str] = None,
```
craffel, line 18: Do we really want to make this optional? Git treats it as non-optional.

git_theta/api.py
```diff
@@ -0,0 +1,71 @@
...
+    else:
+        # Commit directly from python
+        sha = repo.index.commit(commit_msg)
+    return sha
+
+
+# TODO: Should we add something like a `checkout` parameter that actually checks
+# the model out so we are at the right commit after using this model?
+def load(
+    sha: git.Commit,
```
craffel, line 56: Should we try to be fancy and allow loading a file via a tag rather than a commit hash?

git_theta/checkpoints/base.py
```diff
@@ -49,6 +49,15 @@ def load(cls, checkpoint_path):
             should be numpy arrays.
         """
 
+    @classmethod
+    @abstractmethod
+    def from_framework(cls, model_dict):
+        """Convert a checkpoint from the native framework format to git-thetas."""
+
+    @abstractmethod
+    def to_framwork(self):
```
craffel, line 58: framework

git_theta/checkpoints/flax_checkpoint.py
```diff
@@ -16,6 +16,13 @@ class FlaxCheckpoint(Checkpoint):
     def load(cls, checkpoint_path, **kwargs):
         return serialization.msgpack_restore(checkpoint_path.read())
 
+    @classmethod
+    def from_framwork(cls, model_dict):
```
craffel, line 20: framework

git_theta/checkpoints/flax_checkpoint.py
```diff
@@ -16,6 +16,13 @@ class FlaxCheckpoint(Checkpoint):
...
+        return cls(model_dict)
+
+    def to_framwork(self):
```
craffel, line 23: framework

tests/end2end/inprocess/test.sh
```diff
@@ -0,0 +1,28 @@
...
+
+echo "Making model ${MODEL_NAME}"
+INIT_MODEL=`python ../${MODEL_SCRIPT} --action init --seed 1337 --model-name=${MODEL_NAME}`
+
+python test.py
+if [[ "$?" != 0 ]]; then
+    exit 1
+fi
+echo "Verifying that the same model saved in different paths match"
+python ../verify.py --old-model should_match_1.pt --new-model should_match_2.pt
```
craffel, line 21: Is there a reason we can't use the functionality from verify.py directly in test.py rather than doing it separately in a shell script? Or is there a benefit to doing it separately?

git_theta/filters.py
```diff
@@ -0,0 +1,146 @@
+"""Clean and Filter functions."""
```
craffel, line 1: Did these change at all? I'm not inspecting them deeply...

tests/end2end/inprocess/test.py
```diff
@@ -0,0 +1,36 @@
...
+)
+
+print("committing the same model to different paths")
+model_1_sha = git_theta.save(model, "model_1.pt")
+model_2_sha = git_theta.save(model, "model_2.pt")
+print("committing the changed model to the same path.")
+model_1_1_sha = git_theta.save(updated_model, "model_1.pt")
+
+print("Making sure the models we not saved to disk.")
+assert not os.path.exists("model_1.pt")
```
craffel, line 25: Re: your comment above: "TODO: Should we have a `materialize` flag that writes the model to disk so you
# don't have git status showing a diff?" It kind of creeps me out that are referring to things via paths that aren't actually ever materialized on disk...

git_theta/api.py
blester125: We have a default value of `Committing ${path}`, I assume if we make it mandatory there would be a lot of `ups`/`update` commits.

I think the advantage of not having a default is that function sig basically matches torch.save, but we can remove the default if we think it will make people add more informative commits.

git_theta/api.py
blester125: GitPython already supports commit lookup from either a sha1 or a tag so we actually already support that. I've updated argument names to reflect that.

git_theta/checkpoints/base.py
blester125: This explains why the abstractmethod part wasn't working right lol.

Fixed

git_theta/checkpoints/flax_checkpoint.py
blester125: fixed

git_theta/checkpoints/flax_checkpoint.py
blester125: fixed

tests/end2end/inprocess/test.sh
blester125: One reason is that verfiy.py isn't written in a way that makes importing it into other scripts easy, plus doing an import of a file in a different directory without having things setup as an installable package (__init__.py, setup.py, etc) can be weird.

Also the main runner for e2e tests is designed to run shell scripts so we would need to have a special case for this one and duplicate the color prints and things like that.

git_theta/filters.py
blester125: No, they just got moved so they are more reusable.

tests/end2end/inprocess/test.py
blester125: They do get materialized on disk, it's just in the "git future" so they do live on disk, but it's a question of if the git working directory has the right commit checked out. So the path would be materialized in the working tree if you checked out the next commit, but it isn't checked out right now.

I think the main question is just should the checkout step be an optional part of the save/load function (if we force it we would do a bunch of extra I/O we don't need to) or if the caller should just make a checkout afterwards.

git_theta/api.py
craffel: Yeah, I think I am philosophically opposed to making the default to have an uninformative commit message - we want git-theta to be more than torch.save on steroids.

tests/end2end/inprocess/test.py
craffel: Should we have an arg (true by default seems to make sense to me) that specifies whether the checkout should be done after a save/load?

git_theta/api.py
blester125: I made it required

tests/end2end/inprocess/test.py
blester125: I added flags that will keep the on-disk data sync'd to the in-memory model

git_theta/api.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def save(
+    state_dict,
+    path: str,
+    commit_msg: str,
+    checkpoint_type: str = "pytorch",
+    materialize: bool = False,
+) -> git.Commit:
+    """Save a model using git-theta without writing it to disk."""
```
craffel, line 20: Should we make these into proper docstrings if we want people to be calling them from Python?

git_theta/api.py
```diff
@@ -0,0 +1,74 @@
...
+    return sha
+
+
+def load(
+    sha_or_tag: Union[str, git.Commit],
+    path: str,
+    checkpoint_type: str = "pytorch",
+    checkout: bool = False,
+):
+    """Load a model from git-theta without having it checked out."""
```
craffel, line 59: Probably change the docstring given the new checkout arg

tests/end2end/inprocess/test.py
craffel: ... should they default to True?

git_theta/api.py
blester125: Updated

git_theta/api.py
blester125: Updated

tests/end2end/inprocess/test.py
blester125: I don't think they should default to True, this isn't the way we would implement save/load if the main goal is keeping memory and disk in sync. If we wanted to do that our implementations would be:

```python
def save(...):
  ckpt.save(path)
  repo.git.commit(msg)
  
def load(...):
  repo.git.checkout(sha_or_tag)
  ckpt.load(path)
```

When you don't require the working tree has to be in sync with the in-memory then you only need 1 write to git storage for saving and 1 read from git storage for loading. When you force the checkout syncing you add an extra read/write of the model so you end up with the same cost as the simple implementations above (which are really basically wrappers around making/checking out commits, and seems less ergonomic than using git directly).

My thought was that this api would be used for things like, as you are training you make a commit/tag each time your model does better on the dev set and then at the end you load the best performing model and run it on the test set. Since this is a long running program but automatic program we don't care that the working tree is different. Then once the training is done, you can jump around in git history to get different model checkpoints instead of having multiple copies indexed by the step.

craffel: I see. I guess I just worry about someone running a .save and expecting that this implies that the checkpoint is actually saved to disk where they expect it to be (as would happen with a torch.save or whatever). Not sure what the right way to communicate that this is *not* the case.

blester125: I could add a log warning/info that is like 

```
You have saved a model to ${path} directly in git. You are no longer at HEAD, run `git checkout ${sha}` or `git checkout HEAD` to see the model in the working tree.
```

I can see how people might be confused, but just seems weird to have functions that are basically pointless unless non-default fault flags are set lol

craffel: Yeah, I understand the point that these functions are only beneficial/make sense with "False" default, but I think the "False" behavior is probably not what users are expecting from functions called "save"/"load". Is there any way to do the git-to-working-directory copy asynchronously (i.e. not blocking the Python process that did the save/load )or is that a bad idea? Otherwise, maybe it would be clearer if the functions were called "write_to_git"/"load_from_git" or something? Or maybe I am making wrong assumptions about user expectations, not sure.

blester125: Good idea, I renamed them to `save_to_git` and `load_from_git`

craffel (approved):

PR #214, opened 2023-05-04
Repo-level and pattern-level configuration in .thetaconfig

nkandpa2:
This PR factors a bunch of the environment variable-based configuration out into a `.thetaconfig` file at the root of the repository. This file gets committed and distributed as part of the repo.

`.thetaconfig` is a json file with two top-level keys: `repo` and `patterns`. 

The `repo` key maps to a dictionary of repository-level config fields like the atol and rtol parameters for equality checks, the signature size, pool size and threshold for LSH, and the maximum number of concurrent async jobs.

The `patterns` key maps to a list of dictionaries. Each dictionary contains the configuration for a particular glob pattern in the same way that gitattributes are specified for glob patterns. Right now, the only pattern-level configuration is the checkpoint format (PyTorch, TF, etc.).

The `.thetaconfig` file is created whenever a user runs `git theta track` for the first time. Each subsequent `git theta track` command adds another pattern-level dictionary to the file.

Here is an example of what `.thetaconfig` looks like:
```json
{
    "repo": {
        "parameter_atol": 1e-08,
        "parameter_rtol": 1e-05,
        "lsh_signature_size": 16,
        "lsh_threshold": 1e-06,
        "lsh_pool_size": 10000,
        "max_concurrency": -1
    },
    "patterns": [
        {
            "pattern": "model.pt",
            "checkpoint_format": "pytorch"
        }
    ]
}
```

In addition to the code change for creating the `.thetaconfig` file, I also factored the `.gitattributes` reading/writing/parsing out into analogous classes.

Note:
In cases where we want to run a `git` command on a file that matches multiple patterns in `.thetaconfig`, the behavior is that patterns that appear later in `.thetaconfig` shadow previous patterns. This matches the behavior of `.gitattributes`.

For instance say we have the following `.thetaconfig`:
```json
{
    "repo": {
        "parameter_atol": 1e-08,
        "parameter_rtol": 1e-05,
        "lsh_signature_size": 16,
        "lsh_threshold": 1e-06,
        "lsh_pool_size": 10000,
        "max_concurrency": -1
    },
    "patterns": [
        {
            "pattern": "dir1/dir2/model.bin",
            "checkpoint_format": "tf"
        },
        {
            "pattern": "dir1/dir2/**",
            "checkpoint_format": "pytorch"
        }
    ]
}
```
If we run a git command on `dir1/dir2/model.bin`, this would match both patterns, but the second pattern would shadow the first pattern. Thus, the file would be treated as a PyTorch checkpoint. It seems a bit weird since the first pattern exactly matches the file, but this behavior matches Git's parsing behavior for the `.gitattributes` file. Since both `.gitattributes` and `.thetaconfig` are written using the `git theta track` command, we should ensure that the parsing logic for these files is the same.


git_theta/lsh/euclidean_lsh.py
```diff
@@ -68,8 +68,9 @@ def nb_hash(
 
 
 def get_lsh():
-    # TODO we need a better way of keeping track of configuration at the repository level
-    # For LSH configuration, once it is set for a repository, changing it should be handled with care
+    repo = git_utils.get_git_repo()
+    thetaconfig = config.ThetaConfigFile(repo)
```
blester125, line 72: I think we want to make sure that we decouple a lot of these functions from needing there to be an actual config, otherwise they get really hard to reuse or use in tests.

We probably want things like the signature size, atol, etc to be parameters where None values result in looking it up in the config?

git_theta/updates/base.py
```diff
@@ -26,7 +26,8 @@ class Update(metaclass=ABCMeta):
 
     name: str = NotImplemented  # The name used to lookup the plug-in.
 
-    def __init__(self, serializer: params.Serializer, *args, **kwargs):
+    def __init__(self, path: str, serializer: params.Serializer, *args, **kwargs):
```
blester125, line 29: The base class never needs the path right? Can we move this to IncrementalUpdate __init__ like the `update_data` parameter?

PR #215, opened 2023-05-18
Swap Tar Combiner for MsgPack

blester125:
We currently use TAR to combine parameter updates that include more than one tensor into a single blob. This PR updates that blobification to use MsgPack, a binary, json-like format.

Results from a small benchmark for combining 2 `[4096, 4096]` parameters. Note that the combiner is run over a parameter group (for things like partial updates), not over the whole model so this test is fairly representative.

```shell
$ python combiner.py
Size of Tar combined files: 235069440 bytes (235.07MB).
Size of MsgPack combined files: 235031596 bytes (235.03MB).
Saved 37844 bytes (37.84kB).
Tar Time: 14.3261 ± 0.0614 over 50 runs
MsgPack Time: 13.7709 ± 0.2838 over 50 runs
The difference in speed is statistically significant,
p(6.601417392885645e-19<0.05).
```

Size gains are due to Tar including information like file-permissions which are not needed for our use-case (they aren't real files).

Msgpack also yields speed gains when applied to parameter groups of shape `[4096, 20]` which one might see in LoRA.


craffel:
Looks good and I love msgpack, but I'm not sure we need the benchmark code in the repo - any reason you think it's especially important to have in the main source tree?

blester125:
My thought was to build up a collection of micro-benchmarks that inform decisions like this for historical reasons, and to help people select the configuration that works best for them.

Tar vs Msgpack is pretty low-level through so if we don't want to expose this as a configuration option (is only ever use msgpack and delete the tar version) then I don't think we need it.

craffel:
Yeah, I think it's worth documenting, but I've usually put such code in issues/gists rather than in the main source tree just for cleanliness. Not saying it's the right approach.

craffel (approved):

PR #216, opened 2023-05-19
Updates logging to include Async Task Ids.

blester125:
This PR adds new logging handler classes for Stream (console) and File output that can add async task ids to the logging message. This makes debugging eaiser and lets us group logging messages based on parameter name without needing to pass the actual parameter around all the time.

Additionally, the logging configuration has been unified across all our provided scripts and the env var `GIT_THETA_LOG_LEVEL` can be used to control the level of the logs.

Currently, this all still works by configuring the root logger, eventually we should move to a named logger and only configure that one. That will allow us to do things like log to console and file at different levels.

I also changed the default value for `GIT_THETA_UPDATE_PATH` as the default was triggering an extra smudge, even when using dense updates.

### DIFF
--- a/git_theta/updates/dense.py
+++ b/git_theta/updates/dense.py
@@ -21,7 +21,7 @@ class DenseUpdate(Update):
         param_name = "/".join(param_keys)
         logging.debug(f"Reading Dense update for {param_name}")
         tensor = await self.read(param_metadata)
-        logging.debug(f"Read Dense update for {param_name}")
+        logging.debug(f"Finished Read Dense update for {param_name}")
         return tensor
 
     async def write(self, param, param_keys, *args, **kwargs) -> metadata.LfsMetadata:
@@ -29,7 +29,7 @@ class DenseUpdate(Update):
         logging.debug(f"Writing Dense update for {param_name}")
         logging.debug(f"Starting Serializing {param_name}")
         serialized = await self.serializer.serialize({"parameter": param})
-        logging.debug(f"Finsihed Serializing {param_name}")
+        logging.debug(f"Finished Serializing {param_name}")
         logging.debug(f"Starting git-lfs clean for {param_name}")
         lfs_pointer = await git_utils.git_lfs_clean(serialized)
         logging.debug(f"Finished git-lfs clean for {param_name}")
--- a/git_theta/updates/dense.py
+++ b/git_theta/updates/dense.py
@@ -18,11 +18,19 @@ class DenseUpdate(Update):
         super().__init__(*args, **kwargs)
 
     async def apply(self, param_metadata, param_keys, *args, **kwargs) -> Parameter:
-        logging.debug(f"Reading Dense update for {'/'.join(param_keys)}")
-        return await self.read(param_metadata)
+        param_name = "/".join(param_keys)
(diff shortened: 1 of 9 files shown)


git_theta/updates/dense.py
```diff
@@ -18,11 +18,19 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
     async def apply(self, param_metadata, param_keys, *args, **kwargs) -> Parameter:
-        logging.debug(f"Reading Dense update for {'/'.join(param_keys)}")
-        return await self.read(param_metadata)
+        param_name = "/".join(param_keys)
+        logging.debug(f"Reading Dense update for {param_name}")
+        tensor = await self.read(param_metadata)
+        logging.debug(f"Read Dense update for {param_name}")
```
craffel, line 24: Nit: "Finished reading" since "Read" can be read in the present or past tense?

git_theta/updates/dense.py
```diff
@@ -18,11 +18,19 @@ def __init__(self, *args, **kwargs):
...
+        return tensor
 
     async def write(self, param, param_keys, *args, **kwargs) -> metadata.LfsMetadata:
-        logging.debug(f"Writing Dense update for {'/'.join(param_keys)}")
+        param_name = "/".join(param_keys)
+        logging.debug(f"Writing Dense update for {param_name}")
+        logging.debug(f"Starting Serializing {param_name}")
         serialized = await self.serializer.serialize({"parameter": param})
+        logging.debug(f"Finsihed Serializing {param_name}")
```
craffel, line 32: Finished

git_theta/utils.py
```diff
@@ -73,14 +73,15 @@ def __get__(self, obj, objtype=None):
 class EnvVarConstants:
     CHECKPOINT_TYPE = EnvVar(name="GIT_THETA_CHECKPOINT_TYPE", default="pytorch")
     UPDATE_TYPE = EnvVar(name="GIT_THETA_UPDATE_TYPE", default="dense")
-    UPDATE_DATA_PATH = EnvVar(name="GIT_THETA_UPDATE_DATA_PATH", default="update.pt")
+    UPDATE_DATA_PATH = EnvVar(name="GIT_THETA_UPDATE_DATA_PATH", default="")
```
craffel, line 76: Why doesn't this get a default anymore?

craffel (approved):

git_theta/updates/dense.py
blester125: Fixed

git_theta/updates/dense.py
blester125: Fixed

git_theta/utils.py
blester125: Basically, when an update is done with side loaded information (like a sparse or lora update), the on-disk/working checkpoint doesn't change, but the checkout in the git history does (it has the side loaded information in it). This causes false positives when you run `git status`. As such, we include a check for side-loaded information in some of our scripts that will write the updated version to disk so that git doesn't detect a diff.

With the default value, git-theta was always triggering that extra write, even in a dense update where the result is overwriting the model with the same values.


So removing this default value skips that extra smudge in cases where there isn't any side-loaded information.

PR #224, opened 2023-06-08
Add support for the safetensors library.

blester125:
This PR also adds tests!


git_theta/checkpoints/safetensors_checkpoint.py
```diff
@@ -0,0 +1,32 @@
+"""Checkpoint using the HF safetensors format.
+
+safetensors has the ability to write model checkpoint from "dl-native" -> "safetensors"
+and read "safetensors" -> any "dl-native" framework, not just the one that was
+used to write it. Therefore, we read/write with their numpy API.
+"""
+
+from file_or_name import file_or_name
+from safetensors.numpy import load, save
```
craffel, line 9: Can we just `import safetensors.numpy` and call `safetensors.numpy.load`/`safetensors.numpy.load` explicitly below? I think it'd improve readability since we are defining `save` and `load` functions within the class below.

craffel (commented):
Requesting one cosmetic change

craffel (approved):

PR #225, opened 2023-06-08
Feat/win tempfile

blester125:
The PR adds a wrapper around Python's named temporary file
functionality. When used on windows, there are often permission errors
when trying to open the file by name. As a work around we often open the
tempfile so that it is not deleted after leaving the scope, calling
`.flush()` and `.close()` on it, and then
manually deleting it. This encapsulates that workflow.

This is something we that generally only using during testing, however,
but including it in the main library is the easiest way to import it
in tests. I can figure out a way to move it to the test area if we hate
that.

That PR also updates a few tests to use this new function.

### DIFF
--- a/git_theta/utils.py
+++ b/git_theta/utils.py
@@ -1,6 +1,5 @@
 """Utilities for git theta."""
 
-import contextlib
 import dataclasses
 import datetime
 import functools
@@ -8,7 +7,6 @@ import inspect
 import os
 import re
 import subprocess
-import tempfile
 from dataclasses import dataclass
 from enum import Enum
 from types import MethodType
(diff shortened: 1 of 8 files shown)


git_theta/utils.py
```diff
@@ -302,3 +304,30 @@ def touch(path: str):
...
+    if m is not None:
+        raise RuntimeError(
+            f"'mode' argument should not be provided to 'named_temporary_file', got {m}."
+        )
+    d = kwargs.pop("delete", None)
+    if d is not None:
+        raise RuntimeError(
+            f"'delete' argument should not be provided to 'named_temporary_file', got {d}."
+        )
+    with tempfile.NamedTemporaryFile(*args, mode="w", delete=False, **kwargs) as f:
```
craffel, line 329: Maybe I'm misunderstanding, but couldn't you call `named_temporary_file` with positional args that would then cause a argument disagreement? Like, the first positional arg to NamedTemporaryFile is mode, so if I call named_temporary_file("r") I think this line would be providing two different values for the mode argument?

craffel:
It would feel much cleaner to have this out of the base library and in the test dir instead, but it's not the end of the day.

git_theta/utils.py
blester125: Removed the `*args` part, so should still be flexible without that issue.

blester125:
Figured out how to move it out of the main library, but still make it accessible, even when you don't run the tests from the root of the project.

craffel (approved):

Issue #229, opened 2024-02-19
Method to tell if git-theta wasn't installed?


blester125:
There have been a couple of times where I have showed git-theta to new people and I forget to tell them they need to run `git theta install`.

When you forget to install it, `git theta track` works as normal, but when you run `git add` the filters definitions aren't found and git silently just does a normal add.

Would it be possible/make sense to add an install check in the track command or something to avoid it?

craffel:
Yes!

blester125:
One situation that is still challenging is if someone doesn't have git-theta installed and they clone a git-theta repo. Because they don't have it installed, none of our hooks/filters will be setup so their clone/pull will result in them getting a git-theta metadata file on disk instead of a the real checkpoint.

Similarly, if the model file is already tracked, they won't be running `git theta track` so there isn't a chance to detect they haven't installed git theta and their `git add ` will not use our filter and they will end up staging the original model checkpoint. If they run `git theta add` (because they pip installed but never run `git theta install`) we could detect this issue, but that wouldn't follow our "its just like git" messaging if we recommended using `git theta add` over `git add`. This might be an unwinnable situation, but would like to hear anyone's thoughts.

craffel:
Hmm... I can't really think of a way of handling the issue where someone closes a git-theta repo without having git-theta installed. Wouldn't this also be an issue for e.g. git-lfs?

PR #232, opened 2024-02-26
Auto install git-theta when doing a track as people often forget this step

blester125:
related to https://github.com/r-three/git-theta/issues/229


craffel:
Hmm, for some reason I was thinking you'd fail out or warn or prompt the user rather than just doing a transparent install. I'd be a bit worried about this being a hidden side-effect, but I defer to you.

blester125:
Updated, now checking if git theta is installed will check all possible locations for an install (based on the `git config --list` command. So we don't need to worry about how they installed it so we don't have to run a global install each time.

blester125:
closes https://github.com/r-three/git-theta/issues/229

PR #234, opened 2024-03-26
Git Theta Low Memory Mode

blester125:
This PR adds a low memory mode to git theta where some concurrency is sacrificed to keep the memory footprint as low as possible.

The main issues it fixes are:

1) During the clean filter the dl native checkpoint data is piped into the filter by git. It is then passed into the checkpoint loader (for example `torch.load`) and read. This case can cause a transient issue where there is ~2x the model size memory used, the bytes in the stdin buffer and the actual model as tensors. When using `GIT_THETA_LOW_MEMORY=True` the stdin is first written to a temp file before being read from disk.

2) During the parameter cleaning process, there is a map from parameter name to parameter value, as a parameter value is cleaned it is not removed from this map and therefore not garbage collected. This can cause issues as when the parameter is serialized it can case a transient doubling of memory (the tensor itself and the serialized version). This is especially apparent on things like embedding tables. This change removes the parameter values from the map once they are serialized so the memory usage goes down as more of the model is written out.

One future update might be to move from a boolean to a numerical values where something like level 1 low memory just does the checkpoint temp file and higher levels do things like reducing the concurrency to allow for releasing model parameters.

These kind of changes (only loading then releasing subsets of parameters) will be be important if we want to support really big models that have streamable/lazy loading of checkpoints.

There are also a few small changes, `git_theta.py` -> `git_theta_cli.py` as the naming was causing weird import issues (but the cli command is still `git-theta` as it is set by the console script entry point), getting blobs from git was bugged for things like lived in subdirs, and checking in real torch checkpoints that had parameters last on cuda devices was bugged.


blester125:
This also skips some tensorflow tests because of https://github.com/r-three/git-theta/issues/235

Issue #93, opened 2022-11-23
More robust logging configuration


blester125:
Currently logging is only done through the basicConfig and everything is done at the debug level.

We should update this, we should also be logging to a file (whose location is user controllable), especially for clean and smudge filters, we should have some messages at debug and some at info, and a user configurable way to control the verbosity of logs.

Ideally there would also be a way to see our debug messages without getting the ones from GitPython as some of their debug logs look like errors (the message about CYGWIN for example) and like they come from git-theta as we are currently configuring the root logger.

blester125:
Also, now that we are using Async code, we should make something like a custom formatter that injects an asyncio task id into the log. This will make it easy to `grep` out a single async call from the interleaved result you get otherwise.

blester125:
This link has an example of how to get the current async task id which we could add to a custom formatter https://stackoverflow.com/a/53949138

blester125:
https://github.com/r-three/git-theta/pull/216 adds async task ids to the logs, we still should move off the root logger eventually

PR #238, opened 2024-04-19
Move to a named logger.

blester125:
Move to a logger named `git_theta` so that we don't edit the default logger which causes other libraries we use to output all of there logs too.

Now only git-theta logs get output unless other loggers are configured.


blester125:
One clear advantage is that this suppressed the GitPython log message about not being in a CYGWIN environment which multiple people have though was an error in their git-theta install

blester125:
Closes https://github.com/r-three/git-theta/issues/93

Issue #233, opened 2024-02-26
Finer-grained control of `git theta install`


blester125:
Currently `git theta install` only works on the "global" git configuration (generally `~/.gitconfig`). We should expand the functionality to allow it to install git-theta to other locations that git respects. This should probably be done with a new cli argument like `--scope global` which is an enum with values like:

* `--scope system` updates the `/etc/gitconfig` file
* `--scope global` updates the `~/.gitconfig` file (this should be the default). Note: GitPython doesn't respect `$XDG_CONFIG_HOME` environment variable, so I think we can ignore support for that for now,  see https://github.com/r-three/git-theta/issues/122 for more information. We might want to upstream support for this into GitPython.
* `--scope repo` updates the `$GIT_DIR/config` file
* `--scope worktree` updates the `$GIT_DIR/config.worktree` file. Git only looks at this file is some extension is installed, so we might not want to support this one.

The git documentation about these config files can be found at https://git-scm.com/docs/git-config

Once we support these file locations, we should update the "is git theta installed" code from https://github.com/r-three/git-theta/issues/229 to check all these location and throw a warning/crash instead of auto installing git-theta (we don't want to auto install because they now have options of where to install it).

blester125:
Note: The git docs talk about `$XDG_CONFIG_HOME/git/config` with a fallback to `~/.gitconfig` as the "global" option. GitPython seems to have separated these into a `user` and `global` level. We can either follow them with a `--user` scope or a have our `global` scope detect if `$XDG_CONFIG_HOME` is defined and back off to `~/.gitconfig` if it is not.

PR #246, opened 2024-04-25
Allow users to install git-theta in multiple ways

blester125:
There is now a `--scope` argument for `git theta install`. It has the following values:

* `global`: Install git-theta via `~/.gitconfig`
* `user`: Install git-theta via `$XDG_CONFIG/git/config`, this dir needs to exists
* `system`: Install to `/etc/git/config` which often needs root powers
* `repository`: You need to be in a repo and it installs it to `.git/config`

As we use `git config --list` which checking if git-theta is installed, we should find it no matter where it is installed.


blester125:
closes https://github.com/r-three/git-theta/issues/233

Issue #249, opened 2024-04-25
Add retry to end2end tests


blester125:
Probably somewhere here, https://github.com/r-three/git-theta/blob/fef5ef1bf7c81203de0419f8829cda918a1b8f7f/tests/end2end/runner.sh#L21 we should add a retry loop that basically re-runs the clean and test. This will help with the flaky windows tests.

PR #253, opened 2024-05-09
Add retries to the end-2-end tests.

blester125:
This PR adds the ability to do a limited number of retries (3) for end2end tests. Often the windows tests will fail during CI (an error about the event loop being closed) crashing the whole end2end ci pipeline.

As these are flaky, not broken, re-running the failed tests often fixes the problem. But this is annoying to have to get and email and click re-run each time it happens. This new version will rerun failed tests automatically, logging when a test that passed needed multiple tires. Hopefully this makes the windows issue less annoying going forward.

Related to https://github.com/r-three/git-theta/issues/249, need to verify the retries work/make windows more reliable before we close it.

### DIFF
--- a/tests/end2end/runner.sh
+++ b/tests/end2end/runner.sh
@@ -4,21 +4,46 @@ source ./utils.sh
 
 # Create an associated array mapping test name to exit code.
 declare -A TESTS
+# Map testname to the number of times the test was run.
+declare -A RUNS
 
+TRIALS=3
+
+# Each sub-directory in the current directory is a test.
 for test in ./*/
 do
+    # Remove things like dir / from the test name
     testname="${test%*/}"
     testname="${testname#./}"
     yellow_echo "Running Test: ${testname}"
-    echo "============================================================"
+    echo "================================================================="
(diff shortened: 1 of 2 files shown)


blester125:
image

The new retries work, in this run, the smudge test would have crashed the pipeline. Instead it was re-run, worked, and the pipeline passed.

closes https://github.com/r-three/git-theta/issues/249

tests/end2end/runner.sh
```diff
@@ -4,21 +4,46 @@ source ./utils.sh
...
-    else
-        ../clean.sh
-    fi
-    ./test.sh
-    TESTS[${testname}]=${?}
+    # Run the test up to ${TRIALS} times, stopping when it has a return of 0
+    i=0
+    # This is just a non-zero value, just to get us into the loop, it doesn't
+    # matter what the value is as it will be overwritten by the test return code.
+    return_code=2
```
craffel, line 26: Total nit but maybe it would be cleaner to make it an impossible return code like -1

craffel (approved):

PR #254, opened 2024-05-26
Implement the ability to infer the checkpoint type

blester125:
Currently the included sniffers are just based on the file path, but some formats, for example pytorch's pickle and newer zip format, have magic numbers in their headers we can check for in the future.

In order to make the sniffing extensible, it is implemented as a plugin. In order to do that, and to avoid having to import a deep learning framework to do sniffing, the sniffers are separated out into their own files. This lead to the question of why are checkpoint plugs stored in the main repo? If we move them to the actual plugins area we can continue to support them while not needing to install them if a framework isn't going to be used.

An alternative implementation could include the sniffer plugin as part of the main repo, but the usage would be a bit clunkier as we would need to handle import errors (for example a pytorch sniffer might want to use torch but it isn't installed). The current solution would only sniff for checkpoints from frameworks that you have installed.

From a user perspective, this PR results in the following:

Currently, to use git-theta with a framework, say tensorflow, you need to have tensorflow installed in your python environment and you need to tell git theta that a checkpoint is TF via an environment variable. `pip install git-theta[tensorflow]` is provided as an easy way to ensure both git theta and tensorflow are installed, but it doesn't need to be used.

With this version, to use some framework, say tensorflow, you either need to install `git-theta-checkpoints-tensorflow` along with `git-theta` or you can use `pip install git-theta[tensorflow]` as a short cut. If later you want to use git-theta with pytorch, you need to install `git-theta-checkpoint-pytorch`.

I'd like some feedback on if people think this change to how things are installed would be too much for users.


plugins/checkpoints/safetensors/git_theta_checkpoints_safetensors/flycheck_checkpoints.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def to_framework(self):
+        return self
+
+    @classmethod
+    def from_framework(cls, model_dict):
+        return cls(model_dict)
+
+
+def safetensors_sniffer(checkpoint_path: str) -> bool:
```
craffel, line 43: Why is this here and in the sniffer file itself?

